PR #4, opened 2013-09-03
Update OpenCollar - texture.lsl

SLCollarDB:
Fixes error where Menu Buttons and blank items are added to the User_texture= settings / settings dumpcache.

### DIFF
--- a/LSL/OpenCollar - texture.lsl
+++ b/LSL/OpenCollar - texture.lsl
@@ -333,22 +333,28 @@ default
                 else
                 {
                     list lParams = llParseString2List(sStr, [" "], []);
-                    string sElement = llList2String(lParams, 1);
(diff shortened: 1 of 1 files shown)


WendyStarfall:
This would have been a very reasonable fix, a clever idea and would have got it done. My concern with it is if we patch-in protective code, will it allow people unfamiliar with say, javascript, figure out the "why"; in 3.922 we don't have this problem anymore so I will close this issue. Please know that this request was greatly appreciated tho.

PS: Without trying to pour yet more fuel into whatever rivalry there might be in between OpenCollar and CollarDB, it would have been politically awkward and difficult for me to pull this in. Please really don't get this down the wrong throat but what we want to be so very proud of is that the OpenCollar is made by people for people. "SLCollarDB" is even not people, it's a trademark.

Issue #9, opened 2013-09-14
Beta Branch: Parsing defaultsettings on demand


WendyStarfall:
I think we eventually should add a little feature to the Setting menu that would parse a defaultsettings card on clicky. Imagine it like this:
1. wearer/owner creates a custom defaultsettings card
2. plays with stuff, changes cosmetic things and what not
3. wants it all back how it was stored in the card
4. goes to help/debug -> setting and clicks a "Reset" or some such button

Also should we have a llTextBox for users to be able doing defaultsettings copypasta? This would make it possible for people to backup their settings anywhere, a notecard in the inventory, a text or doc file on their computer.. and then just paste them in the textbox to load them up.

ghost:
this also helps the owner. he can add settings by remote

ghost:
Reset to Default button is much appreciated!

JoyStipe:
good luck, I've been wishing it were possible in SL to edit/create notecards inside objects for a long while. I doubt that Nan would've written the original settings script with anything less, and know I wouldn't have pubbed any released edits without that feature if it were possible.

WendyStarfall:
aack I meant to say the /person/ creates it, not the program LOL

WendyStarfall:
Looks like we can scrap the idea on using llTextBox for pasting settings ->

_"The textbox input is limited to 250 bytes (characters)."_

and to be honest... if somebody wants their settings parsed, it's as simple as relogging or taking the gadget on/off; in our current system it's even about backups or designer failsaves; we can come up with new ideas for 4.0

PR #17, opened 2013-09-19
WIP Implement Reset ALL function.

ghost:
I try to chew my teeth into that code. This is a work in progres as i work on my lsl skills. 

The idea for this comit is to implement a ResetALL function to reset the Collar to its default settings. The "reset ALL" button ore better the "Runaway" Button dos not automaticaly reset the Collar. This is expectet behavior so the sub can flee. But the Resetfunction is much needet.

### DIFF
--- a/LSL/OpenCollar - settings.lsl
+++ b/LSL/OpenCollar - settings.lsl
@@ -35,6 +35,7 @@ string DUMPCACHE = "Dump Cache";
 string PREFUSER = "Pref User";
 string PREFDESI = "Pref Desig"; // yes, I hate cutoff buttons
 string WIKI = "Online Guide";
+string RESETALL = "Reset ALL;
 string UPMENU = "^";
 key g_kMenuID;
 key g_kWearer;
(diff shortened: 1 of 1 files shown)


WendyStarfall:
We are already wrapping it up for the main release; the 250 bytes limit for llTextBox I wrote about in issue #9 was disappointing and without this little extra to parse stuff, I kinda feel making an all new feature to parse back user defaults is too exotic. Specific to your situation it would be much less hassle to just make use of the designer defaults. If the wearer ran away, they flip back to pref designer defaults, if so desired, and those settings are back. That way they can have their collars on pref user defaults, alter it in all kind of ways until they flip the switch and if designer and user defaults are stored in the card, permanently so. Joy's guide for the defaults, she wrote it up really well and simple to follow, again, this is all much less hassle than if you try hacking in a new feature to make something happen that's already pretty much in there.

http://code.google.com/p/opencollar/wiki/EvolutionDefaults

#16 I have also been looking into runaway and learned that some users are totally familiar with that chat command, they use it for ridding of owners and that is their preferred way. Other users though are totally familiar with the Reset All button, that is how they know it and did it since forever. We change that now, we just confuse everyone, I cringe at the idea of being asked two dozen times a day "Where is the Reset All button?". We should have renamed that button to Runaway when we changed it from a full (which wasn't necessary) reset to a runaway but now it's too late and everyone is already familiar with the way it even is. I'm sorry and please don't feel rejected or demotivated but I really want to wrap 3.9 up so we can move on to 4.0.0.

No more new features; instead an all new gadget! =)

ghost:
might be to late for the 3.9 version sure. but consider a consistent new approach for 4.0
no disappointment felt. We are here to improve first and foremost ourselves.

Issue #30, opened 2013-09-28
Post button pops up unusable Leash menu


TucoSolo:
**What is your Second Life name?**
Tuco Solo

**What steps will reproduce the problem?**
In the Leash menu, select Post.

**What is the expected output?**
A dialog to select a leashing post.

**What do you see instead?**
The Leash menu (remenu) and shortly after, the expected dialog to select a leashing post. The Leash menu does not respond to any of the buttons.

**What version of the collar (or other item) are you using?**
3.931

**Please provide any additional information below.**

WendyStarfall:
The menu closes if touching any of the buttons which is fine enough for me personally. I've been on the leash and couldn't possibly fix this without breaking stuff elsewhere, see #29 

I say it again here: If we get a kickass, rock solid Real Leash integration with perfect menu overhaul I'm more than grateful and willing to pull it.

KaroWeirsider:
This looks like it requires two changes to OpenCollar - Leash.

The first would be to prevent the Leash Menu from reopening. beginning at line 744:
else if (sVal == "") // no parameters were passed
            {
                DisplayTargetMenu(kMessageID, iAuth, SENSORMODE_FIND_TARGET_FOR_POST_MENU);
            }
would become:
else if (sVal == "") // no parameters were passed
            {
                DisplayTargetMenu(kMessageID, iAuth, SENSORMODE_FIND_TARGET_FOR_POST_MENU);
                g_iReturnMenu = FALSE;
            }

The second change is actually appears to be a separate bug being once you are in the Post selection menu, the "⏏" button does not work (before and after the first change). The fix for this is to change what is now line 932 due to the above insertion:
if (!g_iReturnMenu) return;
becomes
if (g_iReturnMenu) return;

KaroWeirsider:
When testing the rest of the Leash menu buttons, I found some were acting similar to the Post menu button. I understand if you feel it would be best to leave this for Open Collar 4.0.
If you want to implement this in pre 4.0, I believe the following will make everything work. (this supersedes the 
comments made in the previous post)

Add g_iReturnMenu = FALSE; immediately following Line 704 
                DisplayTargetMenu(kMessageID, iAuth, SENSORMODE_FIND_TARGET_FOR_LEASH_MENU);
                g_iReturnMenu = FALSE; //Prevents the Leash Menu from being re-displayed

Add g_iReturnMenu = FALSE; immediately following what is now Line 722
[...]

PR #35, opened 2013-09-30
Update for Issue #30 Post button pops up unusable Leash menu

KaroWeirsider:
Corrections were made so the Leash menu would not reappear when child dialogs were called. It also corrects an issue where the "⏏" button was not functioning in the child menus.


WendyStarfall:
Woot! Thanks a lot and that was quick too!!

Issue #23, opened 2013-09-27
Update button giving 2 dialogs


TucoSolo:
What is your Second Life name?
Tuco Solo

What steps will reproduce the problem?
Rez an updater nearby, hit the Update button

What is the expected output?
A Yes/No-dialog asking me whether I wish to really update.

What do you see instead?
A Yes/No-dialog asking me whether I wish to really update, and a second dialog with the previous dialog (a so called remnu).

What version of the collar (or other item) are you using?
3.929 (and verified with the updated 3.931)

Please provide any additional information below.

TucoSolo:
In the update script, in UserCommand(), at line 233 there's a check for a button "Update" and if so, UserCommand() will be called again. Yet, the llMessageLinked() call at line 242 will also be executed, causing the remenu.
My guess is that the if-statement at line 233 could use a return statement.

for example, current:
        if (submenu == BTN_DO_UPDATE) UserCommand(iNum, "update", id);
change to:
        if (submenu == BTN_DO_UPDATE) return UserCommand(iNum, "update", id);

(unverified)

WendyStarfall:
oh yeah, let's try that! the re-menu for help/debug when you update isn't a problem in that sense but seriously, really annoying and has been around since 3.7

PR #37, opened 2013-09-30
Update for Issue #23 Update button giving 2 dialogs

KaroWeirsider:
TucoSolo's suggested correction is accurate.
Also updated the formatting for consistency

I'm not sure where the URL changes came from. I didn't change them. You may want to verify or correct those.


WendyStarfall:
thanks a lot and I think we got enough now to bump up the version a little :3

PR #42, opened 2013-10-08
bugfixing on Real Leash

RayZopf:
removed unused functions
changed script name to same format as all the others
adapted to latest OC version (modified variables and strings)
changed indent/scripts formal structure
changed format how settings are stored/made defaultsettings NC working
stay in menu after dies/enabling RealLeash


WendyStarfall:
haha wow, tons of work.. thanks a lot; which file is the right one, I'll run a diff and figure it out =)

PR #44, opened 2013-10-10
bugfixes and minor modifications

RayZopf:
realleash:
removed double message on reset script

subspy:
reworked how settings are saved
    - made defaultsettings NC working
    - radar setting "meters" working again, remembers now both settings
added warning messages for wearer on first use
added runaway reset


WendyStarfall:
thanks a ton for those fixes, these are good stuff :D

PR #49, opened 2013-10-25
Notify owners on SubSpy disable

RayZopf:
(only) when safeword is used,
only one message to wearer when owner too


WendyStarfall:
_passes Ray a beer_ \o/

PR #51, opened 2013-11-03
changes to realleash and subspy + workaround for #54

RayZopf:
some more messages to wearer

subspy: cap and discard messages to avoid heap-collision - also send reports out as soon as buffer get's to full


WendyStarfall:
thanks a lot, your commentary on line 32 and 34 in real leash... it's still not a bug, just a caveat <3

PR #67, opened 2013-12-27
subspy: minor bugfix for messages on settings change

RayZopf:
just a workaround so that wearer and owner get at least message on subspy settings change
also show status of options more clearly in subspy addon menu


WendyStarfall:
I like it, thanks Ray!! <3

PR #68, opened 2014-01-10
changes to subspy and hug message code from another repo, new script from North

RayZopf:
heavy work on the subspy to make formatting ugly (to get some working diff)
ported back (subspy): better display if option is on/off, message to wearer+owner when change on feature via menu
minor bugfixes (variable values, message wording)

one commit from another repo

the two additional subspy files in spare_parts could be deleted
the Offline restriction removel.lsl script is done bei North Glenwalker and is worked on


WendyStarfall:
With all respect and love.. no thanks on that one, sorry. =/

PR #70, opened 2014-01-20
subspy formatting, adding in old bugfixes

RayZopf:
maybe pulling changes inworld next time

read previous pull request and commit list ( #69) to understand,
that this just was/is a change in formatting to get a working diff with pre-littlemousy subspy
and to re-add some other bugfixes, order of messsages, message wording, more messages at all


RayZopf:
this incorporates latest littlemousy changes as given in OC main repo, see 979a88

RayZopf:
still not satisfied with the listen() part ... and github not working ....
http://wiki.secondlife.com/wiki/Limits

RayZopf:
after reverting formatting back to littlemousy one, those are the (backported) differences
http://diffchecker.com/duvamg5i
- all changes from pre-littlemousy version are included
- some bugfixes
  and changes to message order
- more messages, esp. on option change

listen() still needs some improvement, see https://github.com/OpenCollar/OpenCollarUpdater/pull/70#issuecomment-32814372

RayZopf:
just to sum up the changes:
- changes to UpdateListener():
  explicitely handle situation when not attached - always turn listener off then
- changes to DialogSpy(): make it absolutly clear what current status of different options is
- change to Notify(): give nearby owner feedback that option was changed
- changes to performSpyCommand(): write message to buffers so other owners get informed (time, place) via report on changes (enabling!)
- change in state_entry(): put some messages back there, so that those ones always come first, before giving feedback if a specific option is enabled
  - currently messages are not in order, giving status info and after that saying ...checking for them...
- listen(): buffer was holding only the amount of chars defined in g_iListenCap before doing a report, changed that ...even if it breaks logic somehow; also there should be no messages bigger than stated in earlier given link to SL message limits
  - g_iListenCap is defined to be close to those values - any message bigger than g_iListenCap / limits is probably faulty and would break something - cap and discard...
  - still listen() needs more work
- changes to link_message(): log option status/status change initiated by settings from defaultsettings notecard, other  OC sources (addons?)
- timer(): moved messages up to state_entry - see comment there
- attach(): only log place when trace is enabled - but then add logline too
- changed(): only log when trace is enabled, only get current place when trace is on
  - g_sLoc should never be empty as it is filled with current place as soon as option is enabled

Issue #79, opened 2014-02-02
Main Menu > Addons > Subspy 


Attitum:
- Maybe Debugging Code that was not removed

Take Basic Collar
Update it (custom) >>>>>>>>>>>  select spy :)
Select ALL 3 Features > Trace Radar Listen
Click Button       > RadarSettings

Observe the Following
[03:50:39] OpenCollar: Update complete!
[03:50:55] OpenCollar: Teleport tracing is now turned on.
[03:50:57] OpenCollar: Avatar radar with range of 4m around oooooTESTooooo Resident is now turned ON.
[03:50:59] OpenCollar: Chat listener enabled.
[03:51:03] OpenCollar: processing button 4
[03:51:03] OpenCollar: processing button 8
[03:51:03] OpenCollar: processing button 18
[03:51:03] OpenCollar: processing button 5
[03:51:03] OpenCollar: processing button 9
[03:51:03] OpenCollar: processing button 15

GUILTY LINE >  llOwnerSay("processing button "+sButtonValue);

PR #81, opened 2014-02-10
< Main Menu > Addons > Subspy  #79

KaroWeirsider:
Corrected the way Debug is handled


WendyStarfall:
Hey Karo, I'm really sorry but I can't take that commit right now, as it clashes with another that fixes more stuff than just debug style and somehow cannot merge it in automatically for some reason. We should discuss the way we debug with the memory implications of debugstatements along the way to 3.960. Thanks a lot for having done this work, I learned something new there!!

RayZopf:
regarding memory consumption:  
I use that debug function in another project. First time I ran into memory issues I commented out all debug lines - and some optimizer completely removed the debug statements and function  
- free memory went from 678 up to 2220
- on a later version (and with some tweaking) 2846 to 6666  
  - 15 debug statements  
  -  2 extensive if (debug) {... multiple ... llOwnerSay}
  - (inlining some functions)
  - (replacing variables useing integer constants)

gain is probably around 1k when commenting out (and seems even better if let delete it)

Issue #80, opened 2014-02-08
script: OpenCollar - hide ,  on approval


ghost:
.....
string ALL = "Collar";  // change from "All"

.....

SetAllElementsAlpha(float fAlpha)
{
    llSetLinkAlpha(LINK_SET, fAlpha, ALL_SIDES);

```
/* ********* remove, do not change alphasettings 
//set alphasettings of all elements to fAlpha (either 1.0 or 0.0 here)
g_lAlphaSettings = [];
integer n;
integer iStop = llGetListLength(g_lElements);
for (n = 0; n < iStop; n++)
{
    string sElement = llList2String(g_lElements, n);
    g_lAlphaSettings += [sElement, Float2String(fAlpha)];
}
*/ 

// *************************************************
// NEW! update alpha for all elements after "Show Collar" (earlier "Show All")
if(fAlpha != 0.0)
{
    integer n;
    integer iStop = llGetListLength(g_lElements);
    for (n = 0; n < iStop; n++)
    {
        string sElement = llList2String(g_lElements, n);        
        integer iIndex = llListFindList(g_lAlphaSettings, [sElement]);
        if(iIndex !=-1)
        {
            float fAlpha = llList2Float(g_lAlphaSettings, iIndex+1);
            SetElementAlpha(sElement,fAlpha);
        }
    }
}    
```

}

PR #85, opened 2014-02-13
Update OpenCollar - hide.lsl

ghost:
total revision, 
added restore settings from the database (by LM_SETTING_REQUEST),
changed menu: 'show / hide All' replaced by 'show/hide collar',

'show/hide collar' now does not affect to hide settings of elements


WendyStarfall:
I will refer to #80 here in the comment. We are looking at it, 3.950 already had a little hack to handle bell/lock/keys dealy when hiding and showing all but this is maybe a better solution? Let's see! :3

WendyStarfall:
This is the right way to go about it. Ромка you rock! :3

PR #90, opened 2014-02-22
Update OpenCollar - anim.lsl

ghost:
must first turn off the AO, and then start the animation


littlemousy:
By switching off the AO before giving another anim to play, is there a chance that the standard linden anim will show briefly before the new one kicks in?

ghost:
hmm... sure You are right
I think close commit

PR #100, opened 2014-03-01
Update OpenCollar - badwords.lsl

ghost:
no need to SETTING_REQUEST at the start script, because settings script sends all settings after rez automatically.
(reduces link_message flow)


littlemousy:
If this script is reset, how does it get it's settings?

ghost:
many scripts is always reset when rez and attach
and 'OpenCollar - settings' sends all settings automatically after 0.5 seconds when rez.
I think there is no need to request settings after reset, to avoid large flow LM_SETTINGS_REQUEST from many scripts and double LM_SETTINGS_RESPOND on start
already have large flow MENUNAME_REQUEST and MENUNAME_RESPONSE, at the start

PR #102, opened 2014-03-01
Update OpenCollar - hide.lsl

ghost:
no need to SETTING_REQUEST at the start script, because settings script sends all settings after rez automatically.
no need to MENUNAME_REQUEST at the start script.
(reduces link_message flow)

### DIFF
--- a/LSL/OpenCollar - hide.lsl
+++ b/LSL/OpenCollar - hide.lsl
@@ -259,22 +259,6 @@ integer AppLocked(key kID)
     }
 }
 
-RequestSettings()
-{
-    // restore settings from DB after reset script
-    llMessageLinked(LINK_SET, LM_SETTING_REQUEST, g_sAppLockToken, NULL_KEY);    
-    //llMessageLinked(LINK_SET, LM_SETTING_REQUEST, g_sScript+"all", NULL_KEY); //reserved for settings w. "all"
-    
-    // request settings for all Elements
-    integer n;
-    integer iStop = llGetListLength(g_lElements);
-    for (n = 0; n < iStop; n++)
-    {
-        string sElement = llList2String(g_lElements, n);
-        llMessageLinked(LINK_SET, LM_SETTING_REQUEST, g_sScript + sElement, NULL_KEY);
-    }
-}
-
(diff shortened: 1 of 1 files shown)


littlemousy:
If this script is reset, how does it get it's settings?

ghost:
many scripts is always reset when rez and attach
and 'OpenCollar - settings' sends settings automatically after 0.5 seconds when rez.
I think there is no need to request settings after reset, to avoid double flow settings on start.

if use  request settings after reset, it is necessary to remove the automatic send all settings in 'OpenCollar - settings' to reduce linkmessage flood, and add option to request all settings for scripts, like 'Grouptoken_All' (this is used in LM_SETTING_DELETE)

littlemousy:
What you say is right, assuming all of the collar scripts are reset together, or when the collar is rezzed from inventory.  I am more concerned over the behaviour when this script alone is reset.

ghost:
reset script alone can when collar unlocked.
then the collar can be detached and attached.
typically, the user does not reset the single script

I try to reduce redundancy events 'inside' collar

WendyStarfall:
This had been tried before and is always something that ends up rolled back. Everyone who becomes interested in the collar sooner or later stumbles upon these resets, changes them and eventually it turns out we need them because SL is acting like a donkey. Right now it's these bizarre and apparently redundant things that make the collar solid and stable and if we mess with it, we won't have a stable release for April. However, we should explore this from 3.970 on in July after the timer and keys are properly in place. These three elements, leash/real leash, timer and keys need to be stable before we attempt things that provoke an API change - when we got there, let's add RLV Touch and RLV View and along the way mess with resets, menus, settings.

tl;dr
You are doing great work there, Romka. The patches make a lot of sense but we should first bring our leash, timer and keys up to speed before doing anything that makes us itch to get into auth and settings.

ghost:
about leash: look at  leash and leashparticle where
else if (iNum == MENUNAME_REQUEST) 
used instead
else if (iNum == MENUNAME_REQUEST && sMessage == PARENTMENU)
and generate excessive MENUNAME_RESPONSE

PR #119, opened 2014-03-16
Update OpenCollar - bell.lsl : submenu, to be or not to be?

ghost:
submenu is not used, and can be safely removed, or reserved for future use

### DIFF
--- a/LSL/OpenCollar - bell.lsl
+++ b/LSL/OpenCollar - bell.lsl
@@ -379,12 +379,7 @@ integer UserCommand(integer iNum, string sStr, key kID) // here iNum: auth value
 {
     if (iNum > COMMAND_WEARER || iNum < COMMAND_OWNER) return FALSE; // sanity check
     string test=llToLower(sStr);
-    if (sStr == "refreshmenu")
-    {
-        //g_lButtons = [];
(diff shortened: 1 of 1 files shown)


littlemousy:
It's not used by any of the other collar code, but its possible there are 3rd party plugins that DO hook into this menu.  It might be a good idea to ask around before removing this.

WendyStarfall:
It's true but I've been wondering - what else could you possibly want your bell to do? It can now be turned on/off, shown/hidden, adjust volume, adjust frequency, change the sound..

EDIT: There could be a Style thing that has bell types like kitty/cow/classic, I guess

(There is also the help menu in the bell that nags me a bit.. the bell and the badwords help, I will tap into that for an uniform help style. Right now the one gives you green text print, the other makes an info dialog.)

ghost:
Change bell style ? it's interesting. how to implement it?
Change sculpts in one prim?
or several different bells with change alpha?

littlemousy:
I think bells should be defined by the collar designer, so adding sculpt maps by script wouldn't be ideal.  If a designer wanted to offer alternative bell shapes, then different bell prims, changing alpha would be good, IMHO.

ghost:
in this case, need different bell script,
or additional plugin to change the alpha, together with the original bell script

PR #129, opened 2014-03-17
Update OpenCollar - menu.lsl  EXIT button ?

ghost:
add EXIT button in Main menu

### DIFF
--- a/LSL/OpenCollar - menu.lsl
+++ b/LSL/OpenCollar - menu.lsl
@@ -68,7 +68,7 @@ integer DIALOG_TIMEOUT = -9002;
 //5000 block is reserved for IM slaves
 
 string UPMENU = "BACK";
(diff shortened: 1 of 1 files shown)


WendyStarfall:
I'm not sure we need an EXIT button because we already have -> http://screencast.com/t/cCB8j8ek
What do others think?

ghost:
oh, yeah, but this button is very small :)
old cat looking for a magnifying glass :)
and yes, what do others think? I'm curious :)

littlemousy:
Yes, the [X] is small... I usually use the ignore button, it does the same thing as the [X]. The only difference between that and an "exit" button would be that exit kills the stride in the dialog list.  Personally, I'd rather keep a clean menu than gain a single stride in a list.

ghost:
well, let the menu will be clean

PR #134, opened 2014-03-19
Update OpenCollar - update.lsl   possible collarversion bug

ghost:
add delete old version if it's saved in db.
to prevent LM_SETTING_RESPONSE conflicts of the new version vs old, if it's saved in settings.


WendyStarfall:
this one was also covered with the new pulls for null keys you sent, thanks!

PR #190, opened 2014-03-27
Update OpenCollar - titler.lsl  fix loses the auth on relog

ghost:
increased to 0.5 default text offset 
add 'last auth' save/restore


WendyStarfall:
thanks! \o/

PR #195, opened 2014-03-31
Update OpenCollar - main.lsl   LOCK UNLOCK ...

ghost:
use constants LOCK UNLOCK instead strings ....
no need to check LOCK UNLOCK in HandleMenuResponse if no more use lock script.

Main and Apps menu handle differ little... just merge it's

Tested in world, it works.


WendyStarfall:
Hey, sorry we are neck-deep into this script yet. You are right that variables are preferable but we need to push regularly or I cannot keep up with steps. This script might get atleast two more pushes.

PR #197, opened 2014-04-02
Update OpenCollar - main.lsl

ghost:
use LOCK UNLOCK constants 
add "Get Updater"
add HandleMenuRemove()


littlemousy:
Thanks Romka, those issues are addressed in another big commit just pushed in.  Mostly already covered, but you spotted one I missed. Thanks for that! Main changes are done now,  so please take a look at the latest version, I'm sure there's more in there I missed.

PR #209, opened 2014-04-05
Update OpenCollar - update.lsl  broken link

ghost:
broken link to marketplace in Get Updater dialog


WendyStarfall:
this is really odd, it works fine in my copy in world but seems it got an extra new line on here, thanks for correcting this

PR #212, opened 2014-04-06
Update OpenCollar - leashParticle.lsl   defaultsettings?

ghost:
added reading default particles parameters from 'defaultsettings' notecard, if it's present.
changed method of save/restore particles settings.
refined treatment menu buttons.

tested on a cat =^_^=

### DIFF
--- a/LSL/OpenCollar - leashParticle.lsl
+++ b/LSL/OpenCollar - leashParticle.lsl
@@ -60,9 +60,14 @@ string L_COLOR      = "Color";
 string L_GRAVITY    = "Gravity";
 string L_SIZE       = "Size";
 string L_LIFE       = "Life";   // ADDED FOR ST
+string L_GLOW       = "Glow";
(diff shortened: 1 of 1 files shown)


JoyStipe:
Glance at Lines 780-782.... I'm wondering: If the value is equal to -3.0, why set it to -3.0? It's already there? I think that this will prevent setting it to -3 _unless_ it's already -3 (redundant). suggestion: change line 780 to:
if (g_vLeashGravity.z < -3)
edit/add= line 789, same reasoning & dealie, but zero instead of -3

add: line 861 (segment) - "if (g_fParticleAge <= 0.5)" send a notify, and that's all? no change of the value? this a menu button click and there's no variable value change if they use .05 (no matter if it's acceptable change or not). Basically, just add in "g_fParticleAge -= 0.5;" above it similar to line 876, should do the trick.

ghost:
as You wish :)   #213
thank you pointed out errors

PR #234, opened 2014-04-20
Update OpenCollar - settings.lsl     prevent script error if notecard absent

ghost:
to prevent script error if notecard absent
added checking defaultsettings notecard

in 'state_entry': loading notecard after getting DESIGN_ID

### DIFF
--- a/LSL/OpenCollar - settings.lsl
+++ b/LSL/OpenCollar - settings.lsl
@@ -704,8 +704,8 @@ default
                     defaultslineid = llGetNotecardLine(defaultscard, defaultsline);
                     card_key = llGetInventoryKey(defaultscard);
                 }
-                llSleep(1.0);   //pause, then send values if inventory changes, in case script was edited and needs its settings again
-                SendValues();
+                //llSleep(1.0);   //pause, then send values if inventory changes, in case script was edited and needs its settings again
+                //SendValues();  // it is not necessary, because it makes after 'notecard EOF'
             }
             else llOwnerSay(defaultscard + " notecard not found!");
         }
--- a/LSL/OpenCollar - settings.lsl
+++ b/LSL/OpenCollar - settings.lsl
@@ -453,7 +453,11 @@ integer UserCommand(integer iNum, string sStr, key kID)
     else if (C == llToLower(LOADCARD))
     {
         defaultsline = 0;
-        defaultslineid = llGetNotecardLine(defaultscard, defaultsline);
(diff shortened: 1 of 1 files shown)


WendyStarfall:
I've been thinking about this pull request for a while now and I honestly believe that in this case a script error would be favourable. The defaultsettings card is an essential part of what makes the collar an OpenCollar and if a third party assembles a collar forgetting this card, they should be made aware of the error the maximum obvious way which would be a script error popping up in the client interface itself.

I believe that a chat message error would be neater but it is very easy to miss. Are the other opinions?

XenHat:
warning, tired logic at work here...

Well, a more obvious way would be to use `llDialog` and warn the user that they require a notecard, then use [llSetScriptState](http://wiki.secondlife.com/wiki/LlSetScriptState) to pause all the other collar scripts until the notecard is found (by detecting inventory change), then unpausing the scripts. This would avoid breaking the collar and allow reset-less, lightweight soft error.

concept logic:

```
// WARNING: This is a logic concept and only serve to express what
// I have in mind at the moment of writing this.
// This IS NOT and SHOULD NOT be considered as a safe script snippet to implement as-is.

togglescripts(int state) //boolean
{
        [insert here for loop to go through every "Opencollar - " script]
        {
            llSetScriptState(i,state); // running
        }
}
changed(integer change)
{
    if ((change & CHANGED_INVENTORY) &&
        (llGetInventoryType("defaultsettings") == INVENTORY_NOTECARD))
    {
        togglescripts(1);
    }
    else
    {
        togglescripts(0);
    }
}
```

XenHat:
In my tired ramblings, I forgot that missing inventory errors doesn't crash the script... Oh well.

XenHat:
bumping since I have something to add... While using the error dialog is, as Wendy points out, an effective way to prevent missing notecards in the collar creation process, I would like to add my 2 cents...
"error: missing 'defaultsettings' notecard" isn't very professional for an error message. I know I tend to over-engineer things, but the added code in this issue could be worked to throw an error still, while wording it like this:
"Error: 'defaultsettings' notecard is missing from the collar. Please add it to ensure proper functionality"

ghost:
in fact, the collar works fine without 'defaultsettings' notecard,
enough to prevent script error

WendyStarfall:
The settings are silent now if the card is missing and I will change the print to inform users can "create" such a card, rather than edit the one in contents. Because this change here became difficult to merge, I asked Sumi to make a similar one for the 3.980 branch which closes this issue. #368 

Thanks a lot again to Romka for bringing this up and letting us think it over!

PR #235, opened 2014-04-21
Update OpenCollar - bell.lsl  little grooming

ghost:
rem unused MESSAGE MAP.
rearrange menu buttons.
add ResetScript on_rez, because all settings are loaded.
556-569: apply bell settings after loading.

### DIFF
--- a/LSL/OpenCollar - bell.lsl
+++ b/LSL/OpenCollar - bell.lsl
@@ -498,7 +498,9 @@ default
     
     on_rez(integer param)
     {
-        llResetScript();
+        g_kWearer=llGetOwner();
+        g_sSubPrefix=AutoPrefix();
+        if (g_iBellOn) llRequestPermissions(g_kWearer,PERMISSION_TAKE_CONTROLS);
     }
     
     link_message(integer iSender, integer iNum, string sStr, key kID)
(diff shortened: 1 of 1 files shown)


littlemousy:
Why do we need to reset the script on rez?  That seems like a bunch of extra work for the server that isn't needed.

ghost:
permission control will resets after reattach or relog, but some variables like g_iHasControl stay TRUE.
and script can't take control again, when settings loaded.

p.s. many scripts are resetts on_rez, why not ?

JoyStipe:
permission_trigger_ ... it is gone on rez but the sound may still be active
within the script. same issue will occur with animations scripts - you'll
notice it in debug box on your viewer if/when it happens

On Tue, Apr 22, 2014 at 1:21 PM, littlemousy notifications@github.comwrote:

> Why do we need to reset the script on rez? That seems like a bunch of
> extra work for the server that isn't needed.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/OpenCollar/OpenCollarUpdater/pull/235#issuecomment-41090677
> .

## 

~ A bug is a feature with seniority

littlemousy:
The code in git already works fine getting permission to take controls, both on login, and on rewear/rez.  What do you mean "script can't take control again, when settings loaded."

ghost:
both variants work fine:
     on_rez(integer param)
     {
         g_kWearer=llGetOwner();
         g_sSubPrefix=AutoPrefix();
         if (g_iBellOn) llRequestPermissions(g_kWearer,PERMISSION_TAKE_CONTROLS);
    }
or 
     on_rez(integer param)
     {
          llResetScript();
     }

we can not use the llResetScript 
p.s. return to first (old) variant

Issue #238, opened 2014-04-28
@versionnew=123456 creates a duplicate reply


george-black:
Bug report from OpenCollar R&D Group Chat:

[05:32] Toy (toy.wylie): Someone who is reading along and has a github account can file it of course. "@versionnew=123456 creates a duplicate reply".

JoyStipe:
if that only happens with that one number, it sounds like an issue for rlv api. the function should work correctly using any positive integer, as detailed on this url:http://wiki.secondlife.com/wiki/RestrainedLifeAPI#Version_Checking

Come to think of it, check yourself for two relays. LOL if you require assistance with that let me know, I have a relay tester that tells me how many relays people are wearing

SilkieSabra:
Ivanneth in world has duplicated this with a different channel number:
@versionnew=70251 also generates a duplicate reply--this needs more testing seems to me.

JoyStipe:
it isn't just that command - verified this issue today (and to more extent)
while testing/debugging an iwo rlv script: gets double responses from relay
(OC built-in relay) to queries: @getoutfit, @getattach, & @getstatus.
Checked/verified only one active relay on at the time.

On Mon, May 19, 2014 at 2:56 PM, SilkieSabra notifications@github.comwrote:

> Ivanneth in world has duplicated this with a different channel number:
> @versionnew=70251 also generates a duplicate reply--this needs more
> testing seems to me.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/OpenCollar/OpenCollarUpdater/issues/238#issuecomment-43563582
> .

## 

~ A bug is a feature with seniority

WendyStarfall:
happens in rlvmain; it's not breaking functionality thankfully so we can continue working on rlvmain more in-depth before having to release anything; shout-out to Satomi and mouse!

littlemousy:
Thanks to Satomi for spotting the bug.  Its in the code that tries to refresh rlv restrictions after teleport.  Swap the >= for a < at line 755, seems to fix it.

PR #246, opened 2014-06-09
Update OpenCollar - rlvmain.lsl

ghost:
I think it will be correctly


littlemousy:
yes, as per https://github.com/OpenCollar/OpenCollarUpdater/issues/238

Unfortunately, this whole "repeat existing restricitons in case they're lost during teleport" section also causes issues with the relay releasing restrictions for specific items of furniture.  Wendy has an alternate fix for that as well, in an in-world submission.

PR #247, opened 2014-06-19
Update OpenCollar - rlvmisc.lsl

NorthGlenwalker:
To add support of "Force Mouselook" using new RLV 2.9 API to rlvmisc

### DIFF
--- a/LSL/OpenCollar - rlvmisc.lsl
+++ b/LSL/OpenCollar - rlvmisc.lsl
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////////////////////////////
 // ------------------------------------------------------------------------------ //
 //                             OpenCollar - rlvmisc                               //
-//                                 version 3.960                                  //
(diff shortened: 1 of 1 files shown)


WendyStarfall:
I still need to test this because I'm not sure if camdistmax/min will work without a value for distance which would be 0 for mouselook, I guess?

Anyway, slut is working on a rlvview bundle which will include stuff like that! :see_no_evil:

NorthGlenwalker:
I tried a distance value and could not get it to work, but it worked ok with no value set.... but yes 2.9 has a number of new intresting extra's

WendyStarfall:
Yes, I agree. There would be a few things in 2.9 that could be used for fun on-the-fly RLV play. Until FS doesn't bump their RLV API to 2.9 though, we still end up with a lot of users asking why it won't work. :unamused:

WendyStarfall:
Let's add in the manual that this feature requires a 2.9 API

Issue #240, opened 2014-05-05
Garbler 


Attitum:
Only 5 Open issues Left - I think we need a new One :)

Ok like always start with your basic collar, update the hell out of it, and add in all relevant  things in this case its GARBLE.
# Issue : 

Sub / Slave / Pet is able to turn off Garble.

Note > Before Ownership Assigned
{ 
[06:38:27] OpenCollar: Your collar is open to the public.
[06:38:39] OpenCollar: oooooTESTooooo Resident ordered you to be quiet
[06:38:54] OpenCollar: Sorry, Ladashah Resident, but only the person who activated the garbler, or one who outranks them, may release the garbler.
}

Note > AFTER Ownership Assigned
{
[06:40:07] OpenCollar: Added oooooTESTooooo Resident to owner.
[06:40:07] OpenCollar: Your owner can have a lot  power over you and you consent to that by making them your owner on your collar. They can leash you, put you in poses, lock your collar, see your location and what you say in local chat.  If you are using RLV they can  undress you, make you wear clothes, restrict your  chat, IMs and TPs as well as force TP you anywhere they like. Please read the help for more info. If you do not consent, you can use the command "lrrunaway" to remove all owners from the collar.
[06:40:16] OpenCollar: Locked.
}
Note > THE SUB HERE TURNED GARBLE OFF   bad bad bad subby
{
[06:40:46] OpenCollar: Ladashah Resident is allowed to talk again
}
# Expected behaviour :

No way in hell subby should be allowed to allow self to yap after its goddess shut it up :)

PR #250, opened 2014-06-26
restrictive garble

WendyStarfall:
proposed fix for #240


WendyStarfall:
The edits in the script need to be attributed also to littlemousy who helped me with the script and logic for the wording.

The pull request itself is the result of a study how we can avoid empty merge commits after syncing a fork with the upstream repo. I will edit the readme during the next few days to explain this.

littlemousy:
</> snafu in auth code test

Issue #241, opened 2014-05-07
3.960 menu issue: menu does not fix when scripts are removed


ghost:
I have expressly left out certain scripts from my pet's OC 3.960. One such script is the Animations script.
When I look at the menu, the Animations option is still there. Going into Options -> Fix Menus does not remove it.

Please ensure your Main script removes scripts that are not present in the OC's inventory, as would be the expected behaviour since earlier versions.

(Closed because I thought this was in the wrong thing, reopened because it was not.)

SilkieSabra:
The main menu is not designed to be a locus for modular elements either to add or subtract them, thus we can't write it to automatically remove or add buttons when you mod the collar.  However you can easily do this manually by removing the string "Animations", from the static buttons list in main, at line 137.

ghost:
This is strange, because in earlier versions I was sure that the main menu did rebuild with removal/addition of scripts. Am I just mistaken or was this changed somewhere along the lines when I did not notice? (Could be, as I usually didn't take out that many scripts.)
It would be nice to see the main menu (re-)made to be more modular as well, considering the problems with things like the Animation system.

In this case, not being a bipedal avatar in SL, the choice was made to drop the Animation system, as well as some others.

WendyStarfall:
Hi, don't worry, the animations system is correctly uninstalled if you unticked it for a custom update. The button stays around but animation specific scripts are removed and scalar animations are no longer increasing the content count in the collar.

In earlier versions the main menu used to arrange itself modular style like all the other menus. We would like to avoid situations where we end up with multi-page main menus, buttons jumping around all over the place, users getting confused after updates by slowly assembling or incomplete main menus in high latency regions.

Animations sticking around as a button is only a cosmetic concern at the moment. The main menu is faster, slimmer and more solid in summary. We can look into having the buttons removed if a "core" feature was uninstalled (RLV, Animations, Leash or whatever is each individuals opinion what counts as "core"..) and also make a decision if those three buttons, which we can now safely offer as free space, should become user customizable, i.e. have Grab there for the leash or #RLV Folders or Pose etc..

ghost:
It's fairly easy to change the menu so that it checks to see if the scripts it is expecting to find are actually there. I could code a patch for it if you need one.

SilkieSabra:
This was a design decision to protect the main menu from changes to its view, not an oversight , so thank you for the offer to write us a patch and may we call on you another time?  It is as i stated quite easy to remove the button in the script itself.

[...]

PR #252, opened 2014-06-27
dynamic main

WendyStarfall:
blanks out Animations and RLV buttons if those features are not
installed; the overall menu buttons stay consistent though and don't
jump around; also Appearance will blank out if the script is /manually/
removed; addresses #241


XenHat:
Oh hey, that's pretty cool.

PR #253, opened 2014-06-27
Multisearch for RLV Folders

WendyStarfall:
This feature is by Medea Destiny:

improved chat functionality for attaching folders. Now accepts two-level
search terms with a /. Example: <prefix>&&gag/rr will find the first
folder with "gag" in the title just as <prefix>&&gag would now, but
instead of attaching that, will look for a subfolder below that with
"rr" in the title, and attach that folder instead.

additional improvement: search with no result caused "no folder found"
message to be sent to kID, which is llGetOwner() in listen event,
changed to g_kAsyncMenuUser to return error message to person issuing
the search command.

### DIFF
--- a/LSL/OpenCollar - rlvfolders.lsl
+++ b/LSL/OpenCollar - rlvfolders.lsl
@@ -1,8 +1,7 @@
-
 ////////////////////////////////////////////////////////////////////////////////////
 // ------------------------------------------------------------------------------ //
 //                            OpenCollar - rlvfolders                             //
-//                                 version 3.960                                  //
+//                                 version 3.961                                  //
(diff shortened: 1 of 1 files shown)


SatomiAhn:
Still trying to understand how it is done.
So basically, we compensate for the weakness of @findfolder by iterated calls to @getinv ?
(Ironically this is how we used to search-and-wear in OC, before @findfolder was invented, or before I realized it was !).

How well does this perform though ?
Is the inherent complexity of this method the reason for all these direct llOwnerSay and local listeners? 

This is a bit sad, in my opinion, after we managed to clean up most scripts from their listen events and all the "dirt" that goes with them (listeners, timers, ...) !

PR #257, opened 2014-06-28
Refactor and Optimize... +...

WendyStarfall:
This refactoring and optimization is by littlemousy:

Refactoring and optimize memory footprint. Fix broken "repeat
restrictions on region change", add "RLV_QUERY" and "RLV_RESPONSE"
messages for checking current rlv state, Add support for tracking
restrictions by script: use textString instead of key. Add StartIM
exception.


WendyStarfall:
Oops forgot to mention, this will take care of #238 :space_invader:

PR #260, opened 2014-06-29
Update OpenCollar Auth- Adds a capture facility

MethodicalHavoc:
Basically this fork of the standard auth script allows you to play out capture scenarios when a non-owner can acquire ownership of the wearer's collar. This function is off by default and can only be turned on by the owner. Varying lengths of capture time are allowed. The capture mode is over-ridden by a standard D/s submission to an owner.

Apart from the capture functions this Auth script should display no differences from the standard 3.960 Auth script. All modifications or additions are marked with /*\* at the end of line in the code for clarity
## Function

To allow for capture roleplay (either permanent or for limited periods) This script adds two additional options to the Access menu of a 3.960 OpenCollar.

1) For wearers:
When wearers are unowned (no names including their own in the owner list), their "✗ Owner" Option (which they won't need in this situation) is replaced by a "Capture Menu" which gives them access to a range of buttons to set up their capture choices. They can increase or reduce their catpure length in this menu in increments of hour, day and week;  they can opt for permanent capture or they can switch capture mode off.

 2) for other users:
When the wearer is in open access mode, they are not currently owned and they have set a capture length or opted for permanent capture; other users will see the "✓ Owner" button which they would not normally be able to use replaced by a "CAPTURE ME!" button. Pressing this button will add them to the owner list of the wearer's collar for the length of time spcified by the wearer in capture settings (indefinitely if they have selected "Permanent"). They are informed of their length of ownership upon capture.

While a captor owns a wearer they have all the usual rights of ownership and the wearer cannot access the capture menu to modify the capture terms.

An owner can add new owners and potentially ownership can change hands as with a normal collar, however ownership will still lapse at the end of the capture period.

When capture elapses all owners are removed from the collar though other settings (including secowners) which may have been changed remain unaffected until changed by the wearer. Upon a capture elapsing the capture setting reverts to its default "Off" state and must be manually set again by the wearer to prevent unwanted immediate recapture.

Even when capture is set, if unowned a wearer can choose to add a conventional owner. If they do this, the ownership does not expire.
## Known issues

1) Subsequent owners are not currently informed of the temporary status of their ownership . I felt this would lead to greater complexity, and impact on script size for little usability gain.

 2) The capture time is notified in seconds rather than hours days and  weeks and years. Once again I felt the user friendliness had to be balanced against greater complexity, and impact on script size.(Currently there are only about 4000 bytes of free space in the script and it needs space for owner lists)

3) Detailed warning of owner's power not presented when opting for capture mode. Uncertain whether it is necessary, someone opting for capture mode would most likely be well aware of what they are setting themselves up for.

4) The timer only operates on RL time (Unix time). This keeps the script compact and seems to be more the way people work short term captures than using SL time.

### DIFF
--- a/LSL/OpenCollar - auth.lsl
+++ b/LSL/OpenCollar - auth.lsl
@@ -20,7 +20,7 @@ integer g_iGroupEnabled = FALSE;
 list g_lSecOwners;//strided list in the form key,name
 list g_lBlackList;//list of blacklisted UUID
 
-string  g_sWikiURL = "http://www.opencollar.at/manual.html";
+string  g_sWikiURL = "http://www.opencollar.at/user-guide.html";
 string g_sParentMenu = "Main";
 string g_sSubMenu = "Access";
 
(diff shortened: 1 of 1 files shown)


WendyStarfall:
Alright, we can solve this as a dedicated capture app making use of new auth roles and the timer service in production. That solves the memory problem for owner lists! ( #308  and #311 )

PR #265, opened 2014-07-02
Update OpenCollar - anim.lsl

NorthGlenwalker:
To show what our current pose is in the pose menu


WendyStarfall:
Good one! -> :beer:

PR #272, opened 2014-07-03
Update OpenCollar - label.lsl   with scroll long text

ghost:
label with scroll for SPARE PARTS


littlemousy:
Nice!  Works fine for me, even with a long label.  I'll be leaving this version in my own collar, just so I can make it scroll a silly message from time to time, but I'll mostly leave it static. 

Good job!

PR #279, opened 2014-07-07
add temporary owners

WendyStarfall:
this is one is by littlemousy; here we get a new owner level which has
the same permission like primary owners with the exception of the
ability to add new owners; we have 15,554 bytes free on start-up which
is still 5,962 bytes more than 3.967 auth used to have; further
explanation and discussion would be desired

### DIFF
--- a/LSL/OpenCollar - auth.lsl
+++ b/LSL/OpenCollar - auth.lsl
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////////////////////////////
 // ------------------------------------------------------------------------------ //
 //                               OpenCollar - auth                                //
-//                                 version 3.968                                  //
+//                                 version 3.969                                  //
 // ------------------------------------------------------------------------------ //
 // Licensed under the GPLv2 with additional requirements specific to Second Life® //
 // and other virtual metaverse environments.  ->  www.opencollar.at/license.html  //
@@ -17,6 +17,7 @@ key g_kWearer;
 list g_lOwners;//strided list in form key,name
 list g_lSecOwners;//strided list in the form key,name
 list g_lBlackList;//list of blacklisted UUID
+list g_lTempOwners;//list of temp owners UUID.  Temp owner is just like normal owner, but can't add new owners.
 
 key g_kGroup = "";
 string g_sGroupName;
@@ -126,6 +127,7 @@ FetchAvi(integer iAuth, string type, string name, key kAv) {
     string out = llDumpList2String(["getavi_", g_sScript, kAv, iAuth, type, name], "|");
     integer i = 0;
     list src = g_lOwners;
+    if (type == "tempowner") src += g_lTempOwners;
     if (type == "secowner") src += g_lSecOwners;
     else if (type == "blacklist") src = g_lBlackList;
     list exclude; // build list of existing-listed keys to exclude from name search
@@ -159,6 +161,7 @@ AuthMenu(key kAv, integer iAuth) {
 RemPersonMenu(key kID, string sToken, integer iAuth) {
     list lPeople;
     if (sToken=="owner") lPeople=g_lOwners;
+    else if (sToken=="tempowner") lPeople=g_lTempOwners;
     else if (sToken=="secowner") lPeople=g_lSecOwners;
     else if (sToken=="blacklist") lPeople=g_lBlackList;
     else return;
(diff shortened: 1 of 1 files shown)


SilkieSabra:
maybe i will bite on this one, since i'm the hardass on changes to the auth code--what makes these owners "temporary" and not just an extra rung on the ladder between sec owner and owner? and could this be done as a plugin?

WendyStarfall:
This would require a plugin to access temporary ownership. OpenCollar stock distribution should come without any such plugin, in my personal opinion because we give others the opportunity to write and distribute their apps, as well as users get a broad choice of their personally desired "temp ownership", be it keys, kidnap RP, capture games and whatever cool ideas app makers come up with.

SilkieSabra:
in that case cool beans!

littlemousy:
Technically, nothing in the auth module makes these owners temporary.  Once added, they will stay for ever until removed.  Having said that, nothing in the auth menus allows you to add or remove them, so its not so bad :)  Some explanation is probably in order.

My aim here was to add somewhere for the collar to store temporary owners, keep them separate from real owners, and keep them in the normal auth system so they will work seamlessly with other plugins.

In this scheme, a temp owner is authenticated in exactly the same way as a regular owner, but has a few restrictions in the auth script.  Namely, they can only change the tempowners list.  This means that if another plugin (for, say, a capture game) can set a temp owner (by saving it to settings), giving that person full owner style rights over the wearer.  They can add other tempowners (in case the game involves trading said captured subs) but cannot interfere with the wearer's normal ownership, meaning owned subs can safely engage in such games too.

This isn't supposed to implement a whole capture system, that is the job of another plugin who's operational parameters I am not worthy to compute.  What it does do is provide an infrastructure in the collar to allow other plugins to offer such facilities without having to change core scripts to do so.

I'd love to hear other people's input on this one, and will be glad to answer any questions you might have.

XenHat:
I haven't looked at the code yet, but we both know you're better than me at
this. Form the written info about it, I agree with this. Especially the
fact that they're separated than the main owners.

ml132:
I really like the idea. Is there any reason to have the add and remove temporary owners button in the dialog menu though? Plugins that use the temporary owners system will be able to add/remove from this list via linked messages and provide buttons for the wearer to add/remove them if desired, so perhaps it's better to keep those buttons under the "Apps" menu to avoid confusing people with stock collars.

WendyStarfall:
The buttons are not apparent in the menu, they are commented out.

NorthGlenwalker:
if we are calling them temporary owners maybe have it so they are not saved with LM_SETTING_SAVE, so they do not survive a relog as well

WendyStarfall:
##### Here are a few requirements for merging something like this:
- For a proper and full implementation, this feature will need a programmer the level of Nandana, Satomi, Joy, Medea or littlemousy herself to be reliably available until completion.
- It requires a similarly sophisticated timer service to live up to it's full potential (for plugins).
- A sample plugin should be provided with documentation how this could be used. Maybe Methodical could make his capture facility (#260) as a light plugin that taps into a possible setup of this kind.
- We need a dedicated and motivated, small group of testers. "Beta" channel won't do here.

I will keep this pull request open for discussion but hope littlemousy will offer a pull herself soon because, as far as I'm able to understand, this is not the variation of the script that should be merged in.

If we cannot fulfill the requirements above, I'm afraid this is a number too big for us at the moment.

SatomiAhn:
I read this thread only quickly.
I wanted to make a couple of observations (without having read proposed code).
- I think (and have been thinking for a long time) that OC needs an intermediary auth level which can lock everything but not change auth settings (i.e. add/remove owners), and I mean in stock auth script, not some addon. This feature is relevant by itself even without having capture addons (but capture addons would use it as a backend).
- Temporary auth does not need to be tied to that intermediary level of auth. It may make sense for capture addons, but, after all, don't temporary bans also make sense? 
- Coming back to my first idea. I would go as far proposing a way to do it. Starting from the observation that we currently have 2 redundant auth levels (secondary and group), I would promote secondary, giving them the ability to lock stuff but not add owners, and keep group as it currently is (only renaming the level to "casual user" maybe). So far it is only trivial changes. One more change is needed: possibility to add "casual" users by name and possibility to have group/open access optionally acquire secondary owner level instead of casual user.
- All of this (except the temporary thing) is roughly what my "new_auth" branch was doing. Maybe have a look. https://github.com/SatomiAhn/ocupdater/tree/new_auth

littlemousy:
Maybe tempowner was a bad choice of term for what I proposed.  Maybe think of it more like "Just like a regular owner but can't change the other auth lists like a regular owner, owner".  This would, however make for some very long variable names.  There is nothing intrinsically temporary about the "temp owners" proposed above, there is no timer mechanism to remove them at any point.

Adding and removing users from any lists could be done using a timer service, which would also make a great addition to the core collar scripts, saving each application that needs one from implementing their own.  I'm thinking this could be used not only for owners, but as part of capture games, collar locking, restrictive poses and RLV functions.

Issue #281, opened 2014-07-08
OpenCollar - customize


WendyStarfall:
#### OpenCollar - customize

Attached to this issue is a pull request for a script, OpenCollar - customize and minor alterations in main and leashParticle for smooth nesting.

Initially based on a merge of appearance with hide it evolved further in the Virtual Disgrace fork. Instead of choosing if to texture, color or hide, we select the element itself and are presented with options to either hide/show and customization styles. Options to position, rotate and re-size are removed fully.

Best explained with an example, let's take a look at a hypothetical element, Spike.

We navigate from the Main Menu to Customize and are presented with:

| Hide/Show | Spike | BACK |
| :-: | :-: | :-: |

Selecting Spike, gives us:

| Hide/Show | Gold | Silver |
| :-: | :-: | :-: |
| **Pink** | **Yellow** | **Green** |
| **Chrome** | **Rust** | **BACK** |

We assume the designer is using a dedicated texture for the Rust and the Chrome style, listing it in our existing notecard textures_custom:

`Spike~Rust,ab12347f-b1d3-46f7-b555-2ae12345678b5`
`Spike~Chrome,ab56789f-b1d3-46f7-b555-2ae12345678b5`

Next we assume that for Gold and Silver styles, she uses a tint on the Chrome texture and also lists a colour vector:

`Spike~Gold,ab56789f-b1d3-46f7-b555-2ae12345678b5,<0.65882, 0.56471, 0.34118>`
`Spike~Silver,ab56789f-b1d3-46f7-b555-2ae12345678b5,<0.70980, 0.77647, 0.81961>`

Finally we assume that styles Pink, Yellow and Green don't use any texture at all:

`Spike~Yellow,,<1.00000, 1.00000, 0.00000>`
`Spike~Pink,,<1.00000, 0.00000, 0.50196>`
`Spike~Green,,<0.00000, 1.00000, 0.00000>`

This take on customization may well be more user friendly as users are not expected to decide method and prior knowledge to creation. In a way, the customization is handed to the designer rather than to the user, with cumbersome menu hopscotch, while still allowing more advanced users to modify in-depth so it fits individually for them.
##### Size/Rotate/Position:

[...]

PR #282, opened 2014-07-08
create customize+...

WendyStarfall:
creates customize script; changes customize in leash options to
particle; nests customize in main instead of appearance; (also does no
longer blank out buttons in main but prints in chat that the script is
not installed in case for animations, rlv and customize) -> #281

### DIFF
--- a/LSL/OpenCollar - customize.lsl
+++ b/LSL/OpenCollar - customize.lsl
@@ -1,0 +1,398 @@
+////////////////////////////////////////////////////////////////////////////////////
+// ------------------------------------------------------------------------------ //
+//                            OpenCollar - customize                              //
+//                                 version 3.960                                  //
+// ------------------------------------------------------------------------------ //
+// Licensed under the GPLv2 with additional requirements specific to Second Life® //
+// and other virtual metaverse environments.  ->  www.opencollar.at/license.html  //
+// ------------------------------------------------------------------------------ //
+// ©   2008 - 2014  Individual Contributors and OpenCollar - submission set free™ //
+// ------------------------------------------------------------------------------ //
+//                    github.com/OpenCollar/OpenCollarUpdater                     //
+// ------------------------------------------------------------------------------ //
+////////////////////////////////////////////////////////////////////////////////////
+
+string g_sSubMenu = "Customize";
+string g_sParentMenu = "Main";
+string CTYPE = "collar";
+key g_kWearer;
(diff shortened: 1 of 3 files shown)


ghost:
I always wondered why in the descriptions of the elements used 'notexture', 'nocolor', 'nohide', if they are not used to customize,  instead 'texture', 'color', 'hide' in the elements that are used to customize?

WendyStarfall:
it grew that way because of the initial thought that "Certain elements should not be hidden", like the base ring of a collar, so we had a ~nohide tag and the rest followed this tradition

meanwhile it would be difficult to change because so many designs depend on the three tags

ghost:
Tags change is not difficult, simple script makes it fast

WendyStarfall:
Yes, I have such a script too and I can and would do that for the 44 designs we have in stock at OpenCollar but I worry other distributors would follow up only very slowly or wouldn't do it at all.

There are currently 3,672 other collar designs on Marketplace where most are OpenCollar scripted and all of those would require physical changes. If we change tags, we must prepare for it well and provide lots of simple and easy to access information. I don't want to decide if this is done or not.

ghost:
elements menu shows ALL elements of the collar without considering notexture and nocolor tags!

I think that it is necessary to make one new tag 'customize' instead 'nocolor' and 'notexture' tags

and build elements list using this 'customize' and 'hide' tags

and make styles without 'binds' to elements, like
Style,texture_key,color,shine

WendyStarfall:
that would be great, right now, it wouldn't show any elements with an empty description, which is a way to solve it and makes it more simple for builders but.. it still kinda sucks

EDIT: We also need a simple way for adding colours, a **very simple and convenient** way. So you can add all your favourite colours to a colour selection..

ghost:
may retain the classic method appearance with texture, color and hide, but combining them into one 'customize' script,
and adding separately 'style' script, using preset styles from 'custom_styles' notecard, and working with 'customize' script by linkmessage?

WendyStarfall:
My understanding of user preference was that most people prefer selecting by element rather than selecting by category. However this is handled tho, we don't really require 4 scripts.

I have a few more ideas that you might like. Writing them down in a bit.

ghost:
preservation of the classical 'appearance' method solves the "elements tags" problem

WendyStarfall:
closing this one, I pushed a customize variant to SPARE PARTS, let's continue discussing in #281

PR #291, opened 2014-07-14
Added support for raw vector as Titler color

XenHat:
This took me a while to figure out... At least it handles potential spaces in vectors fine and is about a quarter of the first attempt.
There is no menu feature, mainly because the color menu is full.

A little explanation.. Bear with me, I'm tired:

At first, I tried to add it as an optional sub-command to titlercolor, however it did create significant bugs and after several additional lines, was still unfixed. I went KISS (Keep It Simple, Stupid) and added a separate command that would handle it fine using 3 lines instead of 12 lines of buggy code.

Feel free to comment if improvement seems required.

### DIFF
--- a/LSL/OpenCollar - titler.lsl
+++ b/LSL/OpenCollar - titler.lsl
@@ -1,7 +1,7 @@
 ////////////////////////////////////////////////////////////////////////////////////
 // ------------------------------------------------------------------------------ //
 //                              OpenCollar - titler                               //
-//                                 version 3.967                                  //
+//                                 version 3.968                                  //
 // ------------------------------------------------------------------------------ //
 // Licensed under the GPLv2 with additional requirements specific to Second Life® //
 // and other virtual metaverse environments.  ->  www.opencollar.at/license.html  //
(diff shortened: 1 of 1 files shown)


WendyStarfall:
Alright, looks like 042a4d629072f89746603ca00a4c8faa2545a12f solves the problem with title colours not sticking so this change would be good to go now. However the 4.0 milestone has "Colour Dialogs" listed and one idea would be to allow users defining their own chart of colours in form of a notecard and eventually in form of textbox and chat command. That would add colours to a dialog to be used by any plugin having to do with colours.

In a way we already have this with the color plugin but it lacks the option of adding new colours without editing the script and is always required to be installed. Before 3.967 even for just a titler. We need to ask ourselves if for such a very basic need, like tinting, if it is elegant to carry around the whole extra bulk of appearance, hide and color.

If we go this way, mouse would integrate this feature probably in our dialog script and we could call this colour dialog from titler, label, appearance or wherever. For your own use, you could list your colours in your personal colour chart and install this in any of your items. Adding a new one via chat or textbox would mean a single dump and you have the whole list updated to stick into the next item.

That's also what is likely going to happen. I personally put my vote for a system that let's designers define customization in a single card, handled by multiple scripts, including titler, only allowing customization pre-defined by the designer unless the user has sufficient skill or patience to learn how they could add their own styles in this card.

A very restrictive way to do it but I like it because the work effort stays with the designer, allowing casual users to magically change their collars with a single touch but teaches more advanced users how to properly configure design options, there for, giving them a nerdier and more rewarding experience and greater sensation of success if they pulled it off.

XenHat:
Understood. I put my faith and support behind the improved colors plugin as part of the 4.0 "branding" (for lack of a better word) as you laid it out above. This seems to be a much better way to handle it.

I'm not sure if it was implied, but making colors globally plug-able would  allow end users to completely strip the bulk out of the collar, further reducing memory footprint.

This patch was done quite hastily to fill a need I ignored was already in the fixing.
I'll close this now.

PR #294, opened 2014-07-21
Limit tracer options to wearer only

Ivanneth:
Only wearer is able to change the tracer settings
Owner gets notified in case of the wearer changing the tracer settings

### DIFF
--- a/LSL/OpenCollar - main.lsl
+++ b/LSL/OpenCollar - main.lsl
@@ -181,19 +181,31 @@ integer UserCommand(integer iNum, string sStr, key kID, integer fromMenu) {
     if (sStr == "menu") MainMenu(kID, iNum);
     else if (sCmd == "trace") {
         //Debug("Doing trace command");
-        if (iNum == COMMAND_OWNER){
+        if (kID == g_kWearer){
             string sValue = llList2String(lParams, 1); 
             if (sValue == "on"){
                 g_iTraceOn=TRUE;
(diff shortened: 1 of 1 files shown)


ghost:
No no no! 
Only owner can and should change tracer settings!
otherwise it is pointless

Ivanneth:
Don't bother yourself with it...It was the solution Littlemousy suggested and Wendy approved in the forum: http://www.opencollar.at/forum.html#!/general:quick-fix-to-get-rid-of-the 
But was rejected there already by Wendy again so no need to worry, trace will stay enforced upon everyone.

WendyStarfall:
Hello Ivanneth. Please try to understand that changes of this kind must always be decided based on majority and community dynamics and I tried to explain that on the forum with virtually just three people in favour of this change.

That is three in several weeks total, versus many asking daily for a more severe espionage style which would be ethically questionable and then result in the situation we had before and yet more asking for a change the other way around or having concerns based on fact rather than melodramatic ideation.

We end up in a loop and alienate a clear majority that is fine with Trace as it is and there is no good reason to spoil an enjoyable fantasy for others just because we ourselves may not swing that way.

The manual page for this feature is an important part of this fantasy and it's popular appeal.

I believe this discussion is a good example to demonstrate how OpenCollar as a kink toy is based on risk-awareness and mutual consent rather than a totalitarian defined rule set of what is deemed safe and sane.

(I will also paste this reply on the forum.)

Issue #301, opened 2014-08-03
Update Leash Effects


NoriMouse:
Now that the new Ribbon Particle effects are implemented and accessible in SL, it could be a good idea to swap out the outdated, disconnected particles with a continuous particle stream. It's as easy as adding PSYS_PART_RIBBON_MASK to the list of particle system flags on line 205 in the script 'OpenCollar - leashParticle'. 

Preliminary discussions regarding the implementation of ribbon leashes into OpenCollar brought up the point that a singular ribbon would look "flat", as ribbon particles do not always face one's camera; however, multiple workarounds may be used to add depth to the leash including the addition of a second particle emitter, rotated by 90 degrees, or by adding an llSetLinkPrimitiveParamsFast(emitter, [PRIM_OMEGA]) call for the leash particle emitter link to make it constantly rotate.

An example of the current leash particles; note the disconnected appearance
![chain old](https://cloud.githubusercontent.com/assets/4810563/3789743/35044a5e-1abe-11e4-937a-22ba59a948f8.png)

An example using a ribbon particle replacement. 
The UUID for a chain ribbon particle texture is 06ffe23a-4819-12b4-dd9b-b4a845d16a57
The UUID for a rope ribbon particle texture is c63b0ed0-948a-6ec5-8119-a66aef9ab5b9
![chainnew](https://cloud.githubusercontent.com/assets/4810563/3789742/35018af8-1abe-11e4-90d4-482bf720fa59.png)

SilkieSabra:
Would the workaround you propose for the flat appearance add to the script load or particle lag?

XenHat:
As far as I could test (with overkill settings, over the course of several
months), the ribbon particle is virtually lag-less (on the server as well, since it's completely viewer-side. Tech talk about it on the LL JIRA points that the server LSL flag is a dummy to prevent compiling error). I don't think that a single particle flag would increase the load time by a significant amount
either.

WendyStarfall:
I've been watching experiments with the ribbon particles and haven't been convinced of their use for chains or round rope. That was a longer while ago though. However, they would be absolutely awesome for silk or cloth leashes and I'd swap my rusty chain for a perfumed silk band anytime. :princess:

This will get mixed reception, some will like it, others will moan and groan, especially if we just dropped oldschool particles and worked in a rotating or dual emitter ribbon leash. My belly tells me, this will be celebrated as option in the Particle sub menu but as a replacement would likely cause a racket.

[...]

PR #302, opened 2014-08-03
#301

NoriMouse:
Proposed Ribbon Particle update to 'OpenCollar - leashParticle' script


WendyStarfall:
will be implemented in motion of leash guideline #341 in form of two new leash particle styles; Band and Wire

Issue #305, opened 2014-08-06
Static poses are kept while moving


Bondmaster:
The static poses set trough the Anim menu are by definition "static".
These might look very odd when a sub need to move, as the effect is to "slide" around while immobile.
I'm submitting a change via pull request that stop the animation (trough the animation standard ease-out, so not always immediate) and set the pose back as soon as the sub stop moving.
It use a timer to check the current AVI movement status, but only if a pose is actually set, so it should not introduce a relevant lag. Also tests didn't show any drawback so far.

WendyStarfall:
There are currently two solutions provided for that:
1. The Sub AO itself
2. The AO Link drop and go script (found in SPARE PARTS)

A partly implemented solution we have in Firestorm internal AOs that makes use of the LSL bridge to simply turn the whole AO off. This could be improved by deactivating to override stands only.

Bondmaster:
I understand the solution you reference works the other way around, stopping the current AO when a pose is set or a collar anim is playing, but don't allow them to temporarily kick back if the anim is moving while the pose is still active and then get the static pose again when the avi stops.

SilkieSabra:
i've often wondered about this because some of the poses stop when the sub walks and others don't, cutie for instance reliably cedes to the walking animation

WendyStarfall:
handled with #364 for the 3.980 branch

We will see if to make selectable walks to avoid duckwalking but so far this works well.

PR #307, opened 2014-08-06
Modded anim.lsl for shifting out of pose when moving

Bondmaster:
Changes for shift out of a set pose when moving/walking (see Issue #305)


WendyStarfall:
this is a good idea and was implemented a bit differently based on llSetAnimationOverride with #364 
we don't need a timer this way and so far, tests show that it also works with Sub AO and FS Internal, what we have to look at yet is how to avoid duckwalking if using FS bridge methods and a user wanted to select a different style of walk

PR #317, opened 2014-08-08
Update OpenCollar - settings.lsl

ghost:
remove unused

### DIFF
--- a/LSL/OpenCollar - settings.lsl
+++ b/LSL/OpenCollar - settings.lsl
@@ -45,7 +45,7 @@ string UPMENU = "BACK";
 key g_kMenuID;
 key g_kWearer;
 string g_sScript;
-string g_sMenuScript="OpenCollar - main"; //for fixmenus
+//string g_sMenuScript="OpenCollar - main"; //for fixmenus, no more use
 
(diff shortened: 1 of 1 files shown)


LSL/OpenCollar - settings.lsl
```diff
@@ -97,19 +97,14 @@ Debug (string str)
     llOwnerSay(llGetScriptName() + ": " + str);
 }
 */
-Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
+Notify(key kAv, string sMsg, integer iAlsoNotifyWearer)
```
ghost, line 100: use llRegionSayTo instead llInstantMessage if recipient in same region

LSL/OpenCollar - settings.lsl
```diff
@@ -628,40 +618,6 @@ default
                 }
                 else Notify(kAv,"Sorry, only Owners & Wearers may acces this feature.",FALSE);
                 
-                //if (sMessage == PREFDESI)
```
ghost, line 631: all moved to  UserCommand, 
there is no need to keep it here

PR #318, opened 2014-08-08
Update OpenCollar - main.lsl   Notify

ghost:


LSL/OpenCollar - main.lsl
```diff
@@ -121,7 +121,8 @@ Dialog(key kID, string sPrompt, list lChoices, list lUtilityButtons, integer iPa
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer) {
```
ghost, line 121: use llRegionSayTo instead llInstantMessage if recipient in same region

PR #319, opened 2014-08-08
Update OpenCollar - anim.lsl  Notify

ghost:


LSL/OpenCollar - anim.lsl
```diff
@@ -119,19 +119,14 @@ string HEIGHTFIX = "HeightFix";
 string g_sHeightFixToken = "HFix";
 integer g_iHeightFix = TRUE;
 
-Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
+Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
```
XenHat, line 122: Extra space?

ghost, line 122: not only

Issue #283, opened 2014-07-08
Do we really want an Options menu?


WendyStarfall:
If we could instead nest something more useful in the Main Menu, like Un/dress and move options into Help/About?

The only actual option in Options would be [] Personal where we give User_ priority over collarid_

XenHat:
Since I always tend to look in Help for stuff like "Update" and what not... Sure, why not?

WendyStarfall:
I will close this because Options could be used for more useful things. Personally, I don't think settings should create this menu but main should and receive "options" from several other scripts.

PR #326, opened 2014-08-08
Update OpenCollar - Notify

ghost:

### DIFF
--- a/LSL/OpenCollar - appearance.lsl
+++ b/LSL/OpenCollar - appearance.lsl
@@ -116,21 +116,16 @@ integer GetOwnerChannel(key kOwner, integer iOffset)
 }
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
 {
-    if (kID == g_kWearer)
+    if (kID == g_kWearer) llOwnerSay(sMsg);
+    else
     {
-        llOwnerSay(sMsg);
-    }
-    else if (llGetAgentSize(kID) != ZERO_VECTOR)
-    {
-        llInstantMessage(kID,sMsg);
-        if (iAlsoNotifyWearer)
+        if (llGetAgentSize(kID) != ZERO_VECTOR)
         {
-            llOwnerSay(sMsg);
+            llRegionSayTo(kID,0,sMsg);
+            llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
         }
-    }
-    else // remote request
-    {
-        llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
+        else llInstantMessage(kID, sMsg);
+        if (iAlsoNotifyWearer) llOwnerSay(sMsg);
     }
 }
 
--- a/LSL/OpenCollar - appearance.lsl
+++ b/LSL/OpenCollar - appearance.lsl
@@ -101,29 +101,12 @@ key Dialog(key kRCPT, string sPrompt, list lChoices, list lUtilityButtons, integ
     return kID;
 } 
 
-integer GetOwnerChannel(key kOwner, integer iOffset)
-{
-    integer iChan = (integer)("0x"+llGetSubString((string)kOwner,2,7)) + iOffset;
-    if (iChan>0)
-    {
-        iChan=iChan*(-1);
-    }
-    if (iChan > -10000)
-    {
-        iChan -= 30000;
-    }
-    return iChan;
-}
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
 {
     if (kID == g_kWearer) llOwnerSay(sMsg);
     else
     {
-        if (llGetAgentSize(kID) != ZERO_VECTOR)
-        {
-            llRegionSayTo(kID,0,sMsg);
-            llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
-        }
+        if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID,0,sMsg);
         else llInstantMessage(kID, sMsg);
         if (iAlsoNotifyWearer) llOwnerSay(sMsg);
     }
--- a/LSL/OpenCollar - timer.lsl
+++ b/LSL/OpenCollar - timer.lsl
@@ -180,23 +180,19 @@ integer GetOwnerChannel(key kOwner, integer iOffset)
     }
     return iChan;
 }
+
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
 {
-    if (kID == g_kWearer)
+    if (kID == g_kWearer) llOwnerSay(sMsg);
+    else
     {
-        llOwnerSay(sMsg);
-    }
-    else if (llGetAgentSize(kID) != ZERO_VECTOR)
-    {
-        llInstantMessage(kID,sMsg);
-        if (iAlsoNotifyWearer)
+        if (llGetAgentSize(kID) != ZERO_VECTOR)
         {
-            llOwnerSay(sMsg);
+            llRegionSayTo(kID,0,sMsg);
+            llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
         }
-    }
-    else // remote request
-    {
-        llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
+        else llInstantMessage(kID, sMsg);
+        if (iAlsoNotifyWearer) llOwnerSay(sMsg);
     }
 }
 
--- a/LSL/OpenCollar - timer.lsl
+++ b/LSL/OpenCollar - timer.lsl
@@ -167,30 +167,13 @@ string PeelToken(string in, integer slot)
     if (!slot) return llGetSubString(in, 0, i);
     return llGetSubString(in, i + 1, -1);
 }
-integer GetOwnerChannel(key kOwner, integer iOffset)
-{
-    integer iChan = (integer)("0x"+llGetSubString((string)kOwner,2,7)) + iOffset;
-    if (iChan>0)
-    {
-        iChan=iChan*(-1);
-    }
-    if (iChan > -10000)
-    {
-        iChan -= 30000;
-    }
-    return iChan;
-}
 
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
 {
     if (kID == g_kWearer) llOwnerSay(sMsg);
     else
     {
-        if (llGetAgentSize(kID) != ZERO_VECTOR)
-        {
-            llRegionSayTo(kID,0,sMsg);
-            llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
-        }
+        if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID,0,sMsg);
         else llInstantMessage(kID, sMsg);
         if (iAlsoNotifyWearer) llOwnerSay(sMsg);
     }
--- a/LSL/OpenCollar - anim.lsl
+++ b/LSL/OpenCollar - anim.lsl
@@ -121,17 +121,12 @@ integer g_iHeightFix = TRUE;
 
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
 {
-    if (kID == g_kWearer)
-    {
-        llOwnerSay(sMsg);
-    }
+    if (kID == g_kWearer) llOwnerSay(sMsg);
     else
     {
-        llInstantMessage(kID, sMsg);
-        if (iAlsoNotifyWearer)
-        {
-            llOwnerSay(sMsg);
-        }
+        if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID,0,sMsg);
+        else llInstantMessage(kID, sMsg);
+        if (iAlsoNotifyWearer) llOwnerSay(sMsg);
     }
 }
 
--- a/LSL/OpenCollar - auth.lsl
+++ b/LSL/OpenCollar - auth.lsl
@@ -102,10 +102,9 @@ string g_sScript;
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer) {
     if (kID == g_kWearer) llOwnerSay(sMsg);
     else {
-        llInstantMessage(kID, sMsg);
-        if (iAlsoNotifyWearer) {
-            llOwnerSay(sMsg);
-        }
+        if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID,0,sMsg);
+        else llInstantMessage(kID, sMsg);
+        if (iAlsoNotifyWearer) llOwnerSay(sMsg);
     }
 }
 
--- a/LSL/OpenCollar - badwords.lsl
+++ b/LSL/OpenCollar - badwords.lsl
@@ -281,17 +281,12 @@ string right(string sSrc, string sDivider) {
 
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
 {
-    if (kID == g_kWearer)
-    {
-        llOwnerSay(sMsg);
-    }
+    if (kID == g_kWearer) llOwnerSay(sMsg);
     else
     {
-        llInstantMessage(kID, sMsg);
-        if (iAlsoNotifyWearer)
-        {
-            llOwnerSay(sMsg);
-        }
+        if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID,0,sMsg);
+        else llInstantMessage(kID, sMsg);
+        if (iAlsoNotifyWearer) llOwnerSay(sMsg);
     }
 }
 
--- a/LSL/OpenCollar - bell.lsl
+++ b/LSL/OpenCollar - bell.lsl
@@ -125,19 +125,15 @@ string AutoPrefix()
 
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
 {
-    if (kID == g_kWearer)
-    {
-        llOwnerSay(sMsg);
-    }
+    if (kID == g_kWearer) llOwnerSay(sMsg);
     else
     {
-        llInstantMessage(kID, sMsg);
-        if (iAlsoNotifyWearer)
-        {
-            llOwnerSay(sMsg);
-        }
+        if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID,0,sMsg);
+        else llInstantMessage(kID, sMsg);
+        if (iAlsoNotifyWearer) llOwnerSay(sMsg);
     }
 }
+
 Debug(string sMsg)
 {
     if (g_iDebugging)
--- a/LSL/OpenCollar - camera.lsl
+++ b/LSL/OpenCollar - camera.lsl
@@ -270,18 +270,13 @@ string TightListTypeDump(list lInput, string sSeperators) {//This function is da
 
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
 {
-    if (kID == g_kWearer)
-    {
-        llOwnerSay(sMsg);
-    }
-    else
-    {
-        llInstantMessage(kID, sMsg);
-        if (iAlsoNotifyWearer)
-        {
-            llOwnerSay(sMsg);
-        }
-    }
+     if (kID == g_kWearer) llOwnerSay(sMsg);
+     else
+     {
+         if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID,0,sMsg);
+         else llInstantMessage(kID, sMsg);
+         if (iAlsoNotifyWearer) llOwnerSay(sMsg);
+     }
 }
 
 Debug(string sStr)
--- a/LSL/OpenCollar - color.lsl
+++ b/LSL/OpenCollar - color.lsl
@@ -130,17 +130,12 @@ key TouchRequest(key kRCPT,  integer iTouchStart, integer iTouchEnd, integer iAu
 
 Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
 {
-    if (kID == g_kWearer)
-    {
-        llOwnerSay(sMsg);
-    }
+    if (kID == g_kWearer) llOwnerSay(sMsg);
     else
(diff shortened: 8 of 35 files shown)


LSL/OpenCollar - appearance.lsl
```diff
@@ -116,21 +116,16 @@ integer GetOwnerChannel(key kOwner, integer iOffset)
...
+    if (kID == g_kWearer) llOwnerSay(sMsg);
+    else
     {
-        llInstantMessage(kID,sMsg);
-        if (iAlsoNotifyWearer)
+        if (llGetAgentSize(kID) != ZERO_VECTOR)
         {
-            llOwnerSay(sMsg);
+            llRegionSayTo(kID,0,sMsg);
+            llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
```
ghost: I'm not sure whether it is: llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);

LSL/OpenCollar - timer.lsl
```diff
@@ -180,23 +180,19 @@ integer GetOwnerChannel(key kOwner, integer iOffset)
...
+    if (kID == g_kWearer) llOwnerSay(sMsg);
+    else
     {
-        llInstantMessage(kID,sMsg);
-        if (iAlsoNotifyWearer)
+        if (llGetAgentSize(kID) != ZERO_VECTOR)
         {
-            llOwnerSay(sMsg);
+            llRegionSayTo(kID,0,sMsg);
+            llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
```
ghost: I'm not sure whether it is: llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);

LSL/OpenCollar - appearance.lsl
littlemousy: I've never been sure about that one, it confused me in the past, so I decided to leave it for another day.
There must be more to it than meets the eye, it doesn't make any sense to me at first sight.

if (llGetAgentSize(kID) != ZERO_VECTOR) //looks like it means "if kID is in the region, use a region say to send the message
but then......
else { llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);} // send them a regionwide message on the owner hud channel?????

I think I'm missing something here, maybe another collar god/dess can bring us up to speed

ghost: I do not see sense too

SatomiAhn:
Maybe the intent was to make the message pass through owner HUD?
But why?

ghost:
I don't know.
what is the point to transmit a message through HUD, if it is possible directly to the owner?

littlemousy:
even if you can pass a message through the owner hud, you can'd do it by llRegionSayTo when it is 
not in your region

ghost:
so I suggest:

Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
{
        if (kID == g_kWearer) llOwnerSay(sMsg);
        else
        {
            if (llGetAgentSize(kID) != ZERO_VECTOR)
              {
                 llRegionSayTo(kID,0,sMsg);
                 llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
            }
           else llInstantMessage(kID, sMsg);
           if (iAlsoNotifyWearer) llOwnerSay(sMsg);
       }
}

or remove llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);

NorthGlenwalker:
There has never been any code in the owners hud to catch this sort of msg, The hud has only ever looked for a pong after sending a ping then it closes the channel.
I can not think of any hud or seperate item OC has ever had that could except messages, which I assume would then be forwarded on to who ever was using the menu at the time.
Unless it was linked to an old grabby post in some way that it reset it's command if it did not have rights to leash the collar wearer, ie it looked for something like "Sorry you can not grab this person" from the collar then closed the grab command.... maybe some very old idea that was never used.

littlemousy:
I'd say remove the GetOwnerChannel line, as no one seems to know what its trying to do.  What's the worst that can happen?  If it breaks something, at least we'll know what it does :)

ghost:
GetOwnerChannel used for llRegionSayTo (kID, GetOwnerChannel (g_kWearer, 1111), sMsg); only,
for anything else.

maybe remove them?

JoyStipe:
GetOwnerChannel is used to specify a unique channel for the wearer's collar, specifically. This prevents "spamming" local or more often, sending a message to several OC wearers in the area (region in this case), as well as being faster. As a channel must be specified for llRegionSayTo, it is preferrable to using local (channel 0).

JoyStipe:
As an else case (wearer not in region), llInstantMessage would be the best method.

ghost:
like this:
if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID,0,sMsg); //recipient in same region
else llInstantMessage(kID, sMsg); //recipient in other region

ghost:
I found out that llRegionSayTo (kID, GetOwnerChannel (g_kWearer, 1111), sMsg) is used to transmit messages to Owner HUD

but why send notify to owner hud?

NorthGlenwalker:
Romka, the Owners hud only ever looks for a "pong" reply when it sends a "ping" from "OpenCollar - listener" when it searches for available collars, so for 99.99% of the time the listen is turned off.
OpenCollar - listener, line 432 - llSay(g_iHUDChan,(string)g_kWearer+":pong");
This should be changed from llSay as well

ghost:
yes, Owners hud only ever looks for a "pong", therefore llRegionSayTo(kID, GetOwnerChannel (g_kWearer, 1111), sMsg) inNotify(key kID, string sMsg, integer iAlsoNotifyWearer) can be removed

JoyStipe:
Just skip this - it appeared that Git never posted it so I "re-wrote" what I think is a better descript on the matter anyways in next post

GetOwnerChannel - see line # 161 of current Listener script for what it is doing. It creates a unique integer, based on wearer's key, and sets that as the special channel for that collar.
Then see line # 282 for the offset (1111 in OpenCollar collars, currently) that this collar will use. The varname may be misleading to you, because when instated, this function was only for communications with OpenCollar's then-brand-new OwnerHUD (it's that old), and we didn't foresee the potential that it could & would come into. That channel is "tuned" to that specific wearer (same number on any collar they wear, as it uses their key), but NOT to any hud (unless that collar's wearer is also wearing a hud). OwnerHuds will not listen openly to all channels, rather only listening for a particular collar's channel when the hud has made a request to it (then listens for response, and stops listening to that channel when it receives a response, or a set number of seconds, whichever comes first).

JoyStipe:
listener script contains the collar's "integer GetOwnerChannel" definition, and the 1111 (offset) deployment.
I will stand and take blame - I wrote that, as well as teh portion of this script that you seem hellbent on wiping out.
why 1111? Arbitrary, totally arbitrary. It can be any number, as long as it is universal across all OpenCollar collars.
What does it actually do? It defines a set integer that will be used as a channel number, for a wearer's collar(s) to communicate with in-world objects.
But what about "g_iHUDChan?" It's a varname. Like any other varname, I selected something that, at that time, made the most sense for it's purpose. Yes it grew beyond being used only for huds, and most of that work was mine as well. Why diudn't I change it then? I couldn't think of a better varname for it, that's all. No grand conspiracy, no special intelligence operation covert conspiracy - I just couldn't think of a better varname.
Why is it used in this script? Because it is the best means to send a message over ONE PRIVATE channel, that we know THAT collar is listening to. Doing this reduces lag time, as opposed to InstantMessage, Whisper, or even RegionSayTo using "local" channel 0. If you can't tell the delay differences, try lag-testing each method at the temple, with a timer set to tick at 0.1 or less seconds.
The results that I discovered multiple times, on multiple sims, in order from longest to briefest -
   Shout, Say, Whisper, RegionSay, local RegionSayTo, and RegionSayTo on a private channel.

So, yes there was a reason for it, and there is the reason. Will it work without that? Sure, but it will add to the "lag" between commands, as well as server-side.

ghost:
llInstantMessage causes the script to sleep for 2.0 seconds. 
if recipient is in the same region, it is better to use llRegionSayTo(kID,0,msg); for Notify() function.
if recipient in other region, need to use llInstantMessage

see notes: http://wiki.secondlife.com/wiki/LlInstantMessage

JoyStipe:
Further clarification (attempt). If I am guessing correctly, you are in the notify-wearer function, at the point of notifying someone who is controlling the collar as it were. IN these cases, most times yeah it's an owner/leasher who is right there. However, this is not always going to be the case. IF the controller is someone with an owner hud, they will have to be in the region, as it currently stands, due to SL constraints, to even issue a request via the hud. If they are using a hud, instant messaging the avi will NOT inform the hud (s)he is using, and some of the returned data will be meant specifically for a hud, making no sense at all to a person. IE - not returning that info to a requesting hud WILL break hud functions

ghost:
casual NorthGlenwalker: HUD listens to all notifications from the collar or only 'pong' response?

https://github.com/OpenCollar/OpenCollarUpdater/pull/326#issuecomment-51692443

it makes no sense to send all notification into Owner HUD channel

WendyStarfall:
I'm so confused..

:bear: "Wendy is confused."
:panda_face: "About the notify?"

No, about the channel 1111.

:panda_face: "This is supposed to reduce lag if sending it over a single channel to the wearer."

Oh, I see!

:bear: "That's right but it is not sent to the collar wearer, it is coming from the collar wearer to someone else, only if this someone is not in the same region."

But how can they receive the command then?

:bear: "They can't. What is done is: _if agent is not in this region, talk to them with a command that only works in this region._"

I'm so confused....

:panda_face: "That's confusing me too but, oh well, if it helps at all, I can explain what happens with Notify here."

Yes, please.

:panda_face: "Alright, these changes introduce a check to see if the person to receive the notify is in the same region. If they are not, we use an IM, if they are we use a region say."

Aha! Why?

:panda_face: "Because the caveat for IM would be to pause the script for two seconds."
:bear: "It's a good change. I would see this as an improvement."
:panda_face: "Yeah."

Cool!
I hope nobody is going to be mad at me when I tell we were going to remove all the Notify stuff anyway.

**_Wendy runs away and hides.**_

:bear: "..."

JoyStipe:
Romka: No. The hud sends request to a collar. Collar sends a response to hud, often using notify routine (since it's there, handy, and doesn't require a duplicate function). In these cases, the OwnerHud is meant to inform the owner when applicable with whatever message (therefore, the collar need not also inform the owner). Most cases (I helped North with dev on this, as well as adapting our collar interfacing for it), the hud will _not_ be relaying info back to the owner.
For example, ping/pong - Hud requests "ping" to a specific wearer/collar, and begins listening on that channel.
Case 1
    \* Collar replies "pong" on channel.
    \* Hud stops listening on that channel, and proceeds to add that wearer's key to a listing for that owner.
Case 2
    \* Collar does not reply (No Auth)
    \* Hud stops listening on that channel when timer kicks (ask North for time amount, as I haven't looked in over a year), and takes no further action.

In this event, the owner need not receive any message from the collar, perhaps not even from the hud. Either way, it is left to the hud to provide notifications to the owner as needed.
Hopefully, this helps to clarify some that a hud will often request data that may even be raw from a collar, which the owner does not need to see, and might even not understand.
In expansion from this: The cuffs project (OpenNC, I believe is what North is calling it) also uses the same method for communications with the collar, if it is still as we established in the past. If so, the wearer would be receiving duplicate notifications from the collar when the cuffs are making requests or replies.
It took months of testing and working to resolve all possible cases to a smoother single-notification method, so please be careful editing the Notify function.

NorthGlenwalker:
Joy has it spot on, the time out in case 2 is 30 seconds as I recall.
The cuffs should never duplicate replys as they work on a seperate channel to to collar (1110) not (1111) else the code is identical ish

JoyStipe:
Wendy, for your "confused" questions.

1111 - Arbitrary offest. It's to prevent every different thing from just using a static integer based only on avi key. the offset provides for the potential of other developers using the same "GetChannel" method winding up with other products using the same identical channels as OC collars. Proven it's value in owner hud. This offset number can be anything 4 digits, but MUST be the same in all OC collars for them to all function correctly with other products meant to interact with them.

used only for wearer outbound messages, not inbound? Nope - it can work both ways, and is in fact the channel that collar will listen to for huds/cuffs/etc making requests. The collar will not listen to anyone else's private OC channel, only it's own channel. See Line #282 for defining of hud channel, then Line #283 Sends to Line #98 (SetListeners), see Line #114, where it tells the collar to listen on that channel.

Outbound usage: please see previous post by myself and North's as well.
Hope that helps overly clarify (or at least utterly confuse hehe)

JoyStipe:
PS - that merge of the Notify "corrections" dumped the whole case set for huds/cuffs/objects wanting responses (all will be broken until it's restored). I will look in my things to see if I have a backup of that. If not, I believe North should have also.

ghost:
I do not understand because of what the dispute.

Only TWO of the all scripts used llRegionSayTo(kID, GetOwnerChannel (g_kWearer, 1111), sMsg) function in Notify(kID....) :  OpenCollar - appearance  and OpenCollar - timer
All others works fine without it.

Any sense?

JoyStipe:
if the channel is pre-defined (in state_entry most often) to a varname, the varname can replace GetOwnerChannel within Notify... I haven't reviewed all of the scripts since this latest merge, and haven't even looked at most of them in a year as it is.

littlemousy:
So, this all hinges on my misunderstanding the purpose of the statement "llGetAgentSize(kID) != ZERO_VECTOR" which in most cases seems to be used to mean "Is avatar in the region".

Here we can be almost certain they are in the region because we normally reply to people that just spoke to us, so must be in the region.

In this case, it means "Is this an avatar", and if false, all comms should be sent back on the ownerhud channel, right?

ghost:
yes, llGetAgentSize(kID) return ZERO_VECTOR if avatar is not in the region or if it is not an avatar.
http://wiki.secondlife.com/wiki/LlGetAgentSize

but no sense send notify to 'owner hud' channel

I think it would be better:

Notify(key kID, string sMsg, integer iAlsoNotifyWearer) {
         if (kID == g_kWearer) llOwnerSay(sMsg); 
         else {
                 if (llGetAgentSize(kID)) llRegionSayTo(kID,0,sMsg);  // avatar in region
                 else llInstantMessage(kID, sMsg);  // avatar not in region
                 if (iAlsoNotifyWearer) llOwnerSay(sMsg);
          }
}

JoyStipe:
After reviewing a few scripts (in particular listener) changes over the  course of the year, all that I'm certain of at this point is that communications in & out for objects (huds, cuffs, aos, etc) are handled by the listener script. Other scripts should be sending anything that is requested by an object back to listener via linked-messages keyed with a special constant integer (`INTERFACE_RESPONSE`).
Note that this will only reply to in-world objects, not to avis, and only under pre-defined cases.
see listener script, `link_message` section, `if (iNum==INTERFACE_RESPONSE)` segment for what is sent back via listener. Any other events that send info to the object must either use Notify or send a linked message keyed with `INTERFACE_RESPONSE`.

As to Notify usage of OwnerChannel, please allow me again to try to clarify WHY that is used. Channel 0, even with a llRegionSayTo is bad. it does have delay, especially on a populated/heavily-scripted sim.
llRegionSayTo, when used with a private channel, has less than 0.03 (might even be under .003, been a while since I tested & I forgot the decimal placement beyond hundredths) seconds delay, on a very heavily populated mainland sim with extreme scripts running.

Using channel 0 (local) with RegionSayTo produced a significant increase in delay (0.4 seconds best time on the same heavily populated sim).

ghost:
okay, 
RegionSayTo with local channel make ~ 0.4 seconds delay.
InstantMessage causes the script to sleep for 2.0 seconds. 
which is better to use if an avatar in the same region?

if we use RegionSayTo with other channel, i.e. HUD channel, then recipient don't hear message without attached Owner HUD. (only if Owner HU forwards the received messages in llOwnerSay, but it don't do it!)

we can't use llSay or llShout with local channel, because all around hear it

WendyStarfall:
Romka spotted this well. He is right.

Please let's find an agreement how this should be handled, so Romka can build a function in all scripts which we can use in one central script to take care of all the happy notify business. Let's stay focused.

JoyStipe:
> else {
> if (llGetAgentSize(kID)) llRegionSayTo(kID,0,sMsg); // avatar in region
> else llInstantMessage(kID, sMsg); // avatar not in region
> if (iAlsoNotifyWearer) llOwnerSay(sMsg);
> }
> 
> which is better to use if an avatar in the same region?

llRegionSayTo is preferable. In way of an ordering, from fastest to slowest:
llRegionSayTo (private), llRegionSayTo (local), llOwnerSay, llInstantMessage, llWhisper, llSay, llShout, llRegionSay.

the problem here is that  the notify should be going to an OBJECT, it won't.
However, from the few scripts that I have glanced at, I don't see any instances where it _was_ set with using OwnerChannel as you described Romka (which would look like)

```
if (llGetAgentSize(kID) == ZERO_VECTOR) llRegionSayTo(kID, GetOwnerChannel(1111), sMsg);
```

or

```
if (llGetAgentSize(kID) != ZERO_VECTOR) llInstantMessage(kID, sMsg);
else llRegionSayTo(kID, GetOwnerChannel(1111), sMsg);
```

 If no script is using that, then no script is sending a notify out for an object to begin with.

ghost:
I'm tired. 
llInstantMessage causes the script to sleep for 2.0 seconds. I try not to use it where possible.
Most similar replacement is llRegionSayTo(kID, PUBLIC_CHANNEL, sMsg);
PUBLIC_CHANNEL or 0. equal.

if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID, 0, sMsg);
else llInstantMessage(kID, sMsg);

or compact variant
if (llGetAgentSize(kID) ) llRegionSayTo(kID, 0, sMsg);
else llInstantMessage(kID, sMsg);

if (llGetAgentSize(kID) != ZERO_VECTOR)
or same 
if (llGetAgentSize(kID) )

JoyStipe:
```
Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
{
    if (kID == g_kWearer) llOwnerSay(sMsg);
    else
    {
        if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID, 0, sMsg);
       else llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
        if (iAlsoNotifyWearer) llOwnerSay(sMsg);
    }
}
```

The above code is streamlined from timer.lsl & can replace all function notify definitions (all scripts that use it), though some scripts will need the integer GetOwnerChannel definition added to them, if so.

There will be an issue (with this) in cases of avatar kID being an avi out of region (such as owner notifications that the wearer took off the collar, etc), which hasn't been resolved yet. This is why the code only appeared in timer & accessory scripts.

JoyStipe:
Looked through the portal list of functions, and the only function I can find that can actually determine whether the passed key is an avi (not in same sim) or not is llRequestAgentData, which would require two key constants + 1 string constant, addition of a dataserver event + timer event for those lovely dataserver errors, and would add more delay time than llInstantMessage.

The below code is the best compromise that I can think up, sending both llInstantMessage (in case kID is an avi out-of-range) and llRegionSayTo with channel (in case kID is an object). This should be sound as a function used by any & all scripts, and able to send subspy messages to the offline avis, as well as responses to huds or cuffs. IF we choose to just eliminate inter-sim & offline owner messages (subspy, rlv turned off, collar unlocked/removed, etc), then we can just eliminate llInstantMessage from Notify functions (beware the wrath of many many users).

```
Notify(key kID, string sMsg, integer iAlsoNotifyWearer)
{
    if (kID == g_kWearer) llOwnerSay(sMsg);
    else
    {
        if (llGetAgentSize(kID) != ZERO_VECTOR) llRegionSayTo(kID, 0, sMsg);
        else
        {
             llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);
             llInstantMessage(kID, sMsg);
        }
        if (iAlsoNotifyWearer) llOwnerSay(sMsg);
    }
}
```

littlemousy:
llGetObjectDetails() returns an empty list if the object is not found.  Could we use this to differentiate between a local object or an absent owner?

JoyStipe:
littlemousy:  In a word, no.
http://wiki.secondlife.com/wiki/LlGetObjectDetails - look at the yellow box and read the text to the right of that... avis who are not in that area would be ignored (including offline), and any object would have to be in the same sim, precluding the possibility of future server enhancement for huds.
The current matter to consider is that without sending InsantMessage out, we defeat subspy and owner notifications of unlocking, removing, rlv status changes, etc. - all such notifications as are normally sent to an owner across the SLobe or even offline.

littlemousy:
We've already identified local avatars with the llGetAgentSize.
We're only left with two possibilities, that's objects, or avatars that are not in the sim.

Earlier you said objects have to be in the same region to communicate, so the problem becomes is it an object in this region, or an avatar that's not in this region.

If llGetObjectDetails returns an empty list, we know that its not an object in this region, so we use the ownerchannel to communicate

otherwise, it must be an offline avatar, so we send an IM

Its true that server changes might break this logic further down the road, but that's true of pretty much anything we write

JoyStipe:
littlemousy, as I mentioned, some notify messages are for avis who are on other sims or even offline. AS for llGetObjectDetails, it will also return empty if an avi key is entered and they are not in range (or offline. I even tested this in-world, to verify. The simple fact is that the function cannot differentiate between an avi or an object, if the uuid called is not in the yellow area (shown in the wiki page, linked above)

ghost:
GetOwnerChannel (g_kWearer, 1111) is used to generate a unique channel to transmit the messages to Owner HUD, but Owner HUD not listen to these messages!
he only listens 'pong' answers to 'ping' request for scan and registration 'Subs' in Owner HUD!

no sense to use llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);

ghost:
By the way, llRegionSayTo(llGetOwner(), channel, message); can be used for a private communication between an avatar attachments. i.e. HUDs, collar, cuffs, e.t.c.

WendyStarfall:
I can't merge this anymore after so many other changes done in the meantime, let's look at it past 3.980 release again, please

Issue #342, opened 2014-08-16
option to allow wearers to unleash themselves


NorthGlenwalker:
I have been asked if we can add an option where an owner can click a button to allow/disallow a wearer to be able to unleash themselves if the owner is not present.
This could stop the problem of a sub from being dragged 90M face first into a wall as their owner logs in.
or any other ideas?

katijaAdored:
i would love to see collar wearers, ,by default, be able to unleash themselves, but to have an actual restriction that allows the owner at any time, to turn off that ability, for as long as they wish. it makes it simple by default, yet leaves the owners in total control. they can use the restriction sometimes or leave it on, like any other restriction. (i.e., if an owner wants a sub to have to stay leashed to a post and be unable to unleash herself, they could set the restriction then turn it back off; or, if they never want them to unleash themselves, they can just leave the restriction on.)

NorthGlenwalker:
also what about an auto unleash if the wearer stops moving towards the owner, IE gets jammed in a wall, so if movement is halted when greater to lenght of leash for more then 5 seconds?

ghost:
instead of an auto-unleash, an auto-TP to the dom's position would work better i believe. No need for the owner to re-leash the sub. its totally transparent and hassle free for both, as there's no need to stop, click the collar, and leash again.

katijaAdored:
i'd like to add another reason it'd be great to be able to unleash by default is that some collars are sold copy/no mod, so you can't go into the leash script to make it so you can unleash. i seriously think allowing for NOT being able to unleash yourself as a restriction and to BE able to by default solves several problems from so many different angles and simplifies everything in several ways.

ghost:
There's also a psychological edge that comes with that option :)
THe dom seeing his/her subbie staying leashed when he/she can detach it anytime feels pretty good :)
Finally, taking more situations into account, i find it a pretty good idea.

ml132:
A problem with the "auto-tp" idea, is that many sims don't allow you to teleport except to the landing zone. On these sims, any teleport commands will just send the sub to the landing zone, resulting in an infinite loop of teleports...

One method that I think would work well, is to disable the leashing if the leash target is ever more than let's say 10m over the leash distance, but only temporarily. The leashing can be automatically reinstated when the sub becomes close to the leash target again.

WendyStarfall:
closes with #345

PR #345, opened 2014-08-17
Update OpenCollar - leash.lsl

KireF:
this tries to fix part of the problems in #342 
the test line 789: iDist<1.2*g_iLength+10)
compares distance between wearer and leasher with a treshold distance that depends on the leash length.
So there are 3 zones
- under the leashlength : the leash is not taut and the wearer  is "free"
- over the leashlength : the leash is taut and the wearer  is pulled
- further (new) : the situation is abnormal and the leash is disabled until the leasher comes back close enough.


WendyStarfall:
relates to #342

PR #350, opened 2014-08-19
Update OpenCollar - hide.lsl  GLOW

ghost:
add save and restore prims glow, before/after hide/show elements and all collar


WendyStarfall:
this one is precious, thank you :icecream:

PR #363, opened 2014-08-23
(Discussion) possible solution to RLV status inconsistency

SumiPerl:


SumiPerl:
Introduces a two second sleep on inventory change.  Unfortunately there isn't a good way to figure out if the change was a script or not - this might be a notecard that was saved for example.

In my testing it seems to solve the need to cycle RLV on/off when individual scripts are recompiled or added.

Not sure if the pros weigh out the cons, and I'm not sure if this might potentially affect anything else, but perhaps worth some consideration

PR #398, opened 2014-09-02
stealth command switch state

SumiPerl:


ghost:
You made а typo, should be Notify(kID,   instead  Notify(iNum,

PR #420, opened 2014-09-06
strict leash mode unsits avatar when leash pulled

marrmist:
prevents groundsitting to avoid being pulled on a leash


WendyStarfall:
I like this one and it suits the option, thank you Marissa!

Issue #434, opened 2014-09-15
Secondary Owner Can't Unleash Following a Rezpost


ryguasu:
This is a report from OpenCollar 3.981...

Suppose you have a secondary owner of someone with initials, e.g., "wr".
# The problem as seen in the chat window

Now do these in sequence:
/1 wrrezpost
/1 wrunleash

The first command works fine, attaching wr to a post.

But the second command fails with "Sorry, someone who outranks you on Whybaby's collar leashed Whybaby already."
# The problem, via GUI menus

Collar Menu > Leash > Post > Rez Post
Collar Menu > Leash > Unleash
# Expected Behavior

This is in contrast to the expected behavior, wherein a secondary owner should always be able to undo any leashing that he/she himself has done.
# Partial Cause Analysis

This is a report from OpenCollar 3.981...

Suppose you have a secondary owner of someone with initials, e.g., "wr".
# The problem as seen in the chat window

Now do these in sequence:
/1 wrrezpost
/1 wrunleash

The first command works fine, attaching wr to a post.

But the second command fails with "Sorry, someone who outranks you on Whybaby's collar leashed Whybaby already."
# The problem, via GUI menus

Collar Menu > Leash > Post > Rez Post
Collar Menu > Leash > Unleash
# Expected Behavior

This is in contrast to the expected behavior, wherein a secondary owner should always be able to undo any leashing that he/she himself has done.
# Root Cause Analysis

The problem seems to be that the primary vs secondary distinction is lost in the interaction between the collar and the Grabby Post that it creates.

If I'm understanding correctly, the interaction works approximately like this for a rezpost:
1. Collar rezzes a GrabbyPost in proximity to the avatar that we're leashing
2. Grabby Post gets a collision_start event
3. It responds by sending a leashto message back to the collar, on the collar's owner channel
4. The collar leashes the avatar to the new post

The problem seems to lie in steps 3 and 4. The Grabby Post does not know whether it was created by primary or secondary order, and of course it doesn't send any primary vs secondary distinction back to the collar. The collar seems to always treat a leashto message on the owner channel as if it pertained to the primary owner.

littlemousy:
Your assumption about the "always treat leashto as owner on the owner channel" isn't correct.  The commands are authed as per the owner of the object sending the command.  

[...]

PR #435, opened 2014-09-15
Fix ownership permissions when doing a rezpost

ryguasu:
The problem addressed here was initially reported as issue #434.

Currently, a rezpost always sets the leash permissions
to COMMAND_OWNER, meaning if anyone but an owner rezposts
a submissive, then that person cannot subsequently
unleash the submissive. (Only the primary owner can unleash them.)
This commit sets the permissions in a rezpost scenario the
same as they would be in any other leashing.

There are other ways to implement this. I went with this because it modified as little code as I could manage.

### DIFF
--- a/LSL/OpenCollar - leash.lsl
+++ b/LSL/OpenCollar - leash.lsl
@@ -116,6 +116,13 @@ integer g_iLeasherInRange=FALSE; //
 integer g_iRLVOn=FALSE;     // To store if RLV was enabled in the collar
 integer g_iAwayCounter=0;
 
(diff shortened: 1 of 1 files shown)


ryguasu:
Ok, I think this is not actually the right approach to my concerns. If the meaning of rezpost were "create a post and attach the submissive to it" (not an unuseful idea), then I like what I'm doing here. But that would be a rather quirky interpretation of what "rezpost" should mean. I might think about an alternate idea. Thanks.

PR #444, opened 2014-09-26
Notify.... again... big change.... new attempt

ghost:
if Wendy does not kill me...

"if (llGetAgentSize(kID))"  is equal  "if (llGetAgentSize(kID) != ZERO_VECTOR)"    then we will brief

"llRegionSayTo(kID, GetOwnerChannel(g_kWearer, 1111), sMsg);" is senseless and removed


WendyStarfall:
I will take this request as soon as I wrapped up one last version before this goes in.

PR #450, opened 2014-10-06
Update OpenCollar - bell.lsl DisplayName

ghost:
user define function to get DisplayName for wearer and toucher.
remove unused tokens
not need spaces for "sort" buttons ON/OFF SHOW/HIDE


WendyStarfall:
Thanks!

:cat: -> :icecream:

PR #470, opened 2014-10-16
Update OpenCollar - appearance.lsl    optimisation

ghost:
remove clutter
slight optimisation
allow Owner access menu when LooksLock

maybe rename script to 'adjust' ?


littlemousy:
Brilliant work, Romka, saved me a lot of time!  Rename would be good, need to talk to management about that :)

PR #491, opened 2014-11-03
Added basic support for Gaeline AOs

Ansariel:
Added basic support for Gaeline AOs (https://marketplace.secondlife.com/stores/35845)

Works if invoked directly from the collar menu, not from the mini ZHAO in this script.

Version number is the same as in the script that can be obtained from the Marketplace.


WendyStarfall:
Thanks a lot!!

PR #503, opened 2014-12-19
Update BUNDLE_99_Clutter_DEPRECATED, remove old rlv scripts

ghost:
remove old unused rlv scripts, they replaced by rlvstuff


WendyStarfall:
Thanks, these adjustment I only made in-world, forgot to push it for a few versions now!

PR #504, opened 2014-12-19
Update OpenCollar - rlvstuff.lsl  SitNow don't work!

ghost:
fix sBehavior for  'sit:<uuid>=force' command


WendyStarfall:
This will break View features on RLV 2.9 (Sit is more important)

PR #505, opened 2014-12-19
Update OpenCollar - listener.lsl   llGetOwnerKey(kID)

ghost:
is no need llGetOwnerKey(kID) in this, is present in auth script,
and  works properly without them, Owner HUD and Grabby/Leash Post too


WendyStarfall:
I remember we had Grabby Post working only briefly in between 3.957 and 3.950 with the HUD at 20 meter range. Making objects able to call the collar, region wide and with teleport commands on public access, is a very bad idea. That's why the Leash Post instead of the Grabby Post.

PR #508, opened 2014-12-21
Update OpenCollar - main.lsl   "Fix Menu" for owners

ghost:
I think it is right.
if owner sees broken menu, he can fix it.


WendyStarfall:
Yes, I agree. Thanks!

PR #532, opened 2015-06-12
Request Permissions twice?

ghost:
in fact, no need to request permission at on_rez
it is done on attach


GarvinTwine:
thanks for the catch Romka ;)

PR #533, opened 2015-06-12
llSleep?  bad idea. I propose to use a timer instead of llSleep.

ghost:
I think that the use llSlep in the init function is a bad idea.
script sleeps for one second at on_rez and can't hanle any events!
but at the same time 'settings' script send all settings, and other scripts can sends linked messages! 
all this events are queued FIFO, but not more than 64 events, all the more are discarded!
and when the script wakes up and asks MENUNAME_REQUEST, he receives spam from FIFO and MENUNAME_RESPONSE from other scripts.
This can cause the "disappearance" of some of the menu buttons and loss of settings.

I propose to use a timer instead of llSleep.


GarvinTwine:
I agree!

Issue #529, opened 2015-04-22
[manual] rlv clear all command missing


KreaSmadga:
Hi,

on the manual the chat command for clear all rlv restrictions is missing ;c)

WendyStarfall:
I hope I can get around to give the website as a whole a facelift for the 4.0 version. Considering to change web-host or maybe go with a free solution all together.

PR #544, opened 2015-07-12
Manual Pages: Card and Links

WendyStarfall:
This includes now a card with a complete and proper list of commands and new "headlines" that act as hyperlinks to the full manual pages.


WendyStarfall:
completes #529 
pages are partly not published yet but the card lists the command

PR #549, opened 2015-07-12
Force unsit when "pull leash"

ghost:
avatar don't follow target when sitting
need force unsit


WendyStarfall:
That's similar to how @marrmist implemented this feature here: https://github.com/OpenCollar/opencollar/commit/2e0f5859fe4926ed089fd4e6ad065005b383b23b

It was then later changed by @littlemousy because a user felt bothered by the unseat command repetitively issued if having rlv debug messages enabled in the viewer, here: https://github.com/OpenCollar/opencollar/commit/6035d422271e1bdc05624059902c82d69b1e6622

The way I see it, the wearer should be unseated at pull if strict leashing is enabled. If strict leashing is disabled, it should require a deliberate yank to unseat the wearer.

This will have to be similarly implemented in the master branch. :+1:

Issue #551, opened 2015-07-14
Misleading notify when wearer sets self as owner twice


WendyStarfall:
Reporter: Eva Ryan

Device: OpenCollar 4.0 I BETA
Version: 20150712.1

Steps:
- Wear the factory-reset device
- Add yourself as owner via chat or menu
- Add yourself as owner via chat or menu again

Result: Could not find any avatars starting with "Firstname Lastname" in this region.

This message happens when:
- the agent is indeed not in the region and
- the agent is already listed with the access role of choice

We also still notify with "Could not find any avatars in this region." when adding someone and they were the last not listed agent in this region. We should get rid of these notifies, they inform about the obvious, worded in a way that makes the user believe something went wrong.

WendyStarfall:
Our thingy happens at line [539 - 549](https://github.com/OpenCollar/opencollar/blob/master/src/collar/oc_dialog.lsl#L539):

``` LSL
numAgents = llGetListLength(agentList);
                if (!numAgents) {
                    string findNotify;
                    if (find != "") findNotify = "starting with \"" + find + "\" ";
                    llInstantMessage(kRCPT, "Could not find any avatars "+ findNotify + "in this region.");
                } else {
                    //Debug("Found avatars:"+llDumpList2String(agentList,","));
                    g_iSelectAviMenu = TRUE;
                    ClearUser(kRCPT);
                    Dialog(kRCPT, "\nChoose the person you like to add:\n", agentList, [UPMENU], 0, kID, -1, iAuth, "getavi_|"+REQ+"|"+TYPE); //iDigits==-1 means dialog should calculate numbered dialogs
                }
```

That stuff was merged from "OpenCollar - getavi.lsl" into oc_dialog.lsl by @littlemousy during the 3.9 API phase sometime in spring 2014. I remember at the time we would discuss if we should keep or drop this message we used to have: "You are the only one in this region. Add yourself?" and eventually did that later down the road.

Suggestion:
- If we keep a notify when an agent couldn't be found, let's notify that an agent couldn't be found.
- If an agent is already on the list, let's notify they are already on the list.

How about that?  :cake:

GarvinTwine:
already working on this (actually done in world)

GarvinTwine:
OK this one got a bit chaotic :( ...
Romka's solution is perfect here, lets take his and revoke mines.
Then i will add my additional changes.

WendyStarfall:
[...]

PR #562, opened 2015-07-16
Misleading notify when wearer sets self as owner twice

ghost:
maybe it solved #551 ?


WendyStarfall:
Looks like you and @GarvinTwine are working on a solution simultaneously.

You have to contact each other and coordinate this in-world.

PR #563, opened 2015-07-16
Issue#557

GarvinTwine:
fixed the issue and added a filter for Display Name


GarvinTwine:
lets try this one again

Issue #569, opened 2015-07-17
Auth and its Settings


WendyStarfall:
In 4.0 version settings we have [a test](https://github.com/OpenCollar/opencollar/blob/master/src/collar/oc_settings.lsl#L344) to prevent garbage entries coming from auth.

This test also has to allow settings like norun, group, public and such so we can load them from card.

GarvinTwine:
should be fixed in #571

PR #571, opened 2015-07-17
allowing other tokens than access lists again #569

GarvinTwine:
removed a too hardcoded check for setting tokens for the auth script to allow settings to be read again other then the 3 access groups


WendyStarfall:
#569

Issue #572, opened 2015-07-18
changed commands to "prefix action command"


GarvinTwine:
We want to get rid of ugly, hard to read and weird to type commands like:
prefixactioncommand
but use:
prefix action command
old examples:
prefixsetpublic           -   prefixunsetpublic
new:
prefix enable public   -   prefix disable public

Also make menus available throughout the collar with commands:
pr menu something

GarvinTwine:
started on this with #573

WendyStarfall:
![3](https://cloud.githubusercontent.com/assets/1100007/8761455/77982f32-2d52-11e5-93d6-aa20afba4bec.png)

Yes! This will make OpenCollar much cooler!

GarvinTwine:
we have to sit down and look though lots of stuff here though...

GarvinTwine:
Started over again with  #577

GarvinTwine:
added rlv commands in #579

GarvinTwine:
added changes in auth script with #580
changed commands to add/remove owner/trust/block:
owner -> add owner
trust -> add trust
block -> add block
removeowner -> remove owner (rm owner) (trust/block)
changed commands that turn on/off:
setgroup -> group on
unsetgroup -> group off
setgroupname -> set groupname
public -> public on
private -> public off
limitrange -> limitrange on
unlimitedrange -> limitrange off

GarvinTwine:
added #582 for titler:
title on
title off
title up
title down
title color <1,1,0>
title text Your text here

titler or menu titler - opens the menu
titler color or menu titler color - opens the color menu
title box - opens the text box

GarvinTwine:
added #583 for bookmarks

GarvinTwine:
added #584 for animations

GarvinTwine:
added #585 for meshlabel:
changed commands to new policy:
label show on
label show off
label text Your text here
label color <1,0,1>
label scroll on
label scroll off

GarvinTwine:
label show on/off seemed weird... so changed it to label on/off in #587

GarvinTwine:
#600 adjustes couples:
changed command 
stopcouples to
stop couples

PR #573, opened 2015-07-18
issue #572 started with com, auth and rlvsys

GarvinTwine:
com needs to TRIM_STRING_HEAD of commands to allow the space between
prefix and following action+command.
changed commands in auth to the new desired model as well as in rlvsys:
prefix add owner Name or ID
prefix remove trust Name or ID

prefix rlv on/off


WendyStarfall:
#572

PR #574, opened 2015-07-19
issue #572 2nd

GarvinTwine:
changed enable.disable to on/off 
fixed a bug i introduced in last commit for rlvsys


GarvinTwine:
rlvsys is somehow wrong version here, not merging this

PR #609, opened 2015-07-28
changed viewer handshake behaviour

GarvinTwine:
the former endless loop to check for a rlv viewer which had to be
stopped by the user themselves by turning rlv in the collar off is now
changed to a default to check for 90 seconds and with no reply from the
viewer to @version the rlvsys script turns rlv off itself.
For convinience a new command was added to change the default of 3 extra
viewer checks to more suitable ones for individuals with very big
inventories or very slow connections so that they need longer and more
checks to verify an rlv viewer. 
the command is:
prefix rlv handshakes [number]


WendyStarfall:
#626

PR #621, opened 2015-08-01
change the order to read collar animation to show them in alpha-numeric

GarvinTwine:
order in the pose menu


WendyStarfall:
#627

Issue #640, opened 2015-08-06
Exceptions have to be reliably set at login, tp and list change


WendyStarfall:
Some user reported that their exceptions were no longer in place after relog on BETA Phase IV.

I could observe a similar behavior on my own 4.0/3.4.7 device but that might very well be down to me working on it and not to a fault in the program. After I logged on, my exceptions were not in place.

Nevertheless this has to be closely observed by as many beta testers as possible, taking variables like network, bandwidth and inventory sizes into account, in case this has to do with script startup order.

Please observe and study active rlv restrictions in the viewer on relog, teleport and access list change.

PR #648, opened 2015-08-07
for issue #640

GarvinTwine:
removed(commented out) all exception related code to see how this works
:)


GarvinTwine:
Oh, also changed the settings token finally from rlvmain to rlvsys here ;)

Issue #659, opened 2015-08-09
Leash not active when Follow is On


MystyBoots:
Maybe this is intentional, but I'm not sure so I'll report it and I'll also provide a recommendation if it is intentional.

Setup:  Person A has kidnap On.  Person B does a kidnap so "follow" is automatically turned on.

Action:  Person B access collar > leash > Grab.  However, the leash particles will not show and it appears Person A is not leashed but remains only on Follow.  If Person B turns off follow and then grabs the leash, the leash appears and the person appears leashed.

Related:  If Person A is leashed and Person B turns on Follow, follow is not activated while leashed.  However, there's no message appearing in chat.

Recommended fix:  If a person is on Follow and the captor tried to leash them, maybe have text message that reads, "Your subject is already on Follow.  If you wish to leash them, turn off Follow and then grab leash."  And for the related issue above say, "This person is already leashed.  If you wish them to follow you without the leash, unleash them and then activate follow."

WendyStarfall:
I can see how it is irritating if that happens but I don't like the idea of handling this with a notify. It seems a bit strange to me that the program expects the user to solve the problem. We can try to test for leash state and in case we already do follow or are grabbed, switch to the desired new state.

MystyBoots:
That would be a great fix!  Automatically switch to the desired one....   You're Awesome!

GarvinTwine:
fixed in  #665

PR #665, opened 2015-08-10
fix for issue #659

GarvinTwine:
removed a check to prevent releashing while already leashed which also
prevented to change leash states (follow to leash and viseversa)


WendyStarfall:
#659

PR #695, opened 2015-09-24
division by zero

ghost:
if (g_iItemCounter == g_iTotalItems) g_iTotalItems= 0;  - ????

in line 84:  fCount = 100*(fCount/g_iTotalItems); - when g_iTotalItems = 0  (division by zero) it causes error message "OpenCollar Updater - 150917.1 [script:oc_installer_bundles] Script run-time error Math Error"


GarvinTwine:
I think that this is a typo...
it should be:
if (g_iItemCounter == g_iTotalItems) **g_iItemCounter** = 0; 
but i think this doesnt really matter anymore since the script is restarted after a complete installation.
So removing this line should be fine, too.

PR #702, opened 2015-10-04
use NameURI in AccessList menu.

ghost:
I think that unnecessary to use 2-stride lists in Auth settings, if we use NameURI.

Use only UUID in lists saves memory

It might be worth it to change owner trust block lists form?

### DIFF
--- a/src/collar/oc_auth.lsl
+++ b/src/collar/oc_auth.lsl
@@ -456,26 +456,26 @@ UserCommand(integer iNum, string sStr, key kID, integer iRemenu) { // here iNum:
             integer iLength = llGetListLength(g_lOwner);
             string sOutput="";
             while (iLength)
-                sOutput += "\n" + llList2String(g_lOwner, --iLength) + " (" + llList2String(g_lOwner,  --iLength) + ")";
+                sOutput += "\n" + NameURI(llList2String(g_lOwner, iLength-=2));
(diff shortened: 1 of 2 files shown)


XenHat:
Just adding my 2 cents here: I appreciate that. It makes the script more efficient AND it allows the user to see the names like they want (since the viewer formats agent URLs using the user preferences)

For quick reference: `/about` opens web profile, `/inspect` opens the little toast, `/im` opens an IM session. A lot can be done with those special urls. :smile:

GarvinTwine:
Sorry guys...
If we do want to release one day we have to stop to make changes, and this is a real change.
I thought about changing the whole auth lists to 1 stride due to the URI use but put that back to a later release as it is a major change in the auth script and there has to be a time to stop adding changes.
As long as the auth script itself uses the Username and not the UUID to delete from the lists and the command to remove someone from a auth list is also with the Username and is known for 7 years. That is why i think it should be consistent and also have that Name shown in the list for deleting, else i only see confusion here.
We show the names as URI in the menu already, when we print out the list, we do exactly that here, we print the UUID and Username, that what is in the list.

We did a **lot** with these URIs in 4.0 xen, i think this is the only place where you see a avatar name not displayed like that.

Dont get me wrong, the change is good but i really think this should be done completely with the auth lists to UUIDs only and we are at a Release Candidate point right now where such a major change should not be added anymore.
I totally agree we should do this, for 4.1 :)

PR #704, opened 2015-10-07
a little easier with the g_sDistCard

ghost:
Why use a loop to search for notecard?
object inventory can't contain some notecards with same name.

I'm curious what the uuid "4da2b231-87e1-45e4-a067-05cf3a5027ea"...


WendyStarfall:
The UUID is for a special card we can use for distributors to verify their origin. You and others who contributed or who wish to distribute will receive this card and then the "origin" cannot be spoofed.

Issue #698, opened 2015-09-24
Some thoughts about ~nocolor ~notexture ~nohide  e.t.c.


ghost:
may be better "invert" description?
write ~сolor~texture~hide (e.t.c.)   only in those prims where it is necessary,
instead of ~nocolor~notexture~noshiny~noglow~nohide in all others?

WendyStarfall:
I've thought the same very often and it just itches you in the fingers to change that, doesn't it?

In previous API versions this would have been problematic with updates, so it never was actually done.

Anyone who might be wondering what this is about:

Right now we only have to give an element a ~no... tag if this element should be customized in i.e. color but no texture. In that sense, as long as an element is not "named" at all, we don't have to worry about tags in descriptions. However, if looking at the creative process itself, the way to handle this is non-ideal. I rather would decide "And this piece, I want to be able to color" so add a ~color tag to it. The way it currently is handled, works backwards and the process would have to be "And this piece, I don't want to be able to texture but I want to be able to color it" ~notexture and little Miss collar-maker gets a headache.

tl;dr - now is a great chance to finally make this change

RomkaSwallowtail:
if you're interested, I have a ready script:
https://github.com/RomkaSwallowtail/opencollar/blob/master/src/collar/oc_disgraced_themes.lsl

RomkaSwallowtail:
if you're interested, I have a ready script:
https://github.com/RomkaSwallowtail/opencollar/blob/invert-descriptions/src/collar/oc_disgraced_themes.lsl

and description "converter":
https://github.com/RomkaSwallowtail/opencollar/blob/invert-descriptions/src/spares/Convert_Descriptions.lsl

PR #705, opened 2015-10-08
new way of prims descriptions

ghost:
use ~texture ~color ~shiny ~glow ~hide in prims description  for #698


WendyStarfall:
Unfortunately I already began building version 4.0 yesterday with old descriptions after there was no clear majority in favour of new prim descriptions for customizable elements. I think most people already got used to the old way and this is much more apparent to us because of the way we look at the script.

4.0 will continue to use the `~notexture~nocolor~noshiny~noglow~nohide` way. I wished we had done this sooner but with the release candidate already built in-world we are a bit late. Sorry :( :icecream:

ghost:
You can use https://github.com/RomkaSwallowtail/opencollar/blob/patch-1/src/spares/oc_customizer.lsl for automatic change elements descriptions

PR #711, opened 2015-10-11
Update oc_relay.lsl

ghost:
"OpenCollar 3.9" > "OpenCollar 4.0"


WendyStarfall:
This string is like a signature. Changing it to 4.0 would achieve the exact opposite of what we want:
- **To retire and to no longer support this kind of device-internal relay with OpenCollar 4.0**

I'm not necessarily feeling happy about the fact that right before release we have the history flooded with commits to the relay when we don't even want to feature it but I understand that some users will want to stick this script into their 4.0 collars anyway and that's fine.

That doesn't make it an OpenCollar 4.0 implementation in my opinion.

PR #713, opened 2015-10-16
rework for 4.0 api

ghost:
cagehome
garble
timer
relay
rlvstuff
undress


WendyStarfall:
**_If you are doing things this way git won't be tracking authorship and previous logs._** :unamused: While I agree the --follow hack isn't normally as important, those scripts are really just that, artifacts for archaeologists.

EDIT: I removed the crabby commentary and want to apologize because I see you want to do this for good reasons but please try to put yourself in my position. I want to release version 4.0 next week and not really worry about 10 obsolete scripts. We put those in the foreground, and just like at 3.9 with R&D vendors, users will be asking about old stuff instead of really looking at the extremely good and very hard work we did these last two years.

I'm not going to take anymore PRs until after release.

ghost:
well, I will be doing file by file

GarvinTwine:
I do not think it is a really good idea to just make 3.9 spare-parts 4.0 functional...
This should be a good moment to simply let go of old "trash" and start new without carrying real old, hardly reviewed code into another api version.

PR #715, opened 2015-10-16
new cagehome

ghost:
based on OpenCollar - cagehome
reworked for collar 4.0


ghost:
let go of old "trash"

PR #716, opened 2015-10-16
new customizer

ghost:
based on OpenCollar - customizer
reworked for collar 4.0


ghost:
let go of old "trash"

PR #718, opened 2015-10-16
Update oc_undress.lsl

ghost:
reworked fo collar 4.0


ghost:
let go of old "trash"

PR #720, opened 2015-10-19
oc_meshlabel.lsl -  double redraw label when rez/attach

ghost:
to prevent double redraw label at rez/attach

### DIFF
--- a/src/collar/oc_meshlabel.lsl
+++ b/src/collar/oc_meshlabel.lsl
@@ -351,7 +351,7 @@ default
             llMessageLinked(LINK_SET, MENUNAME_REMOVE, g_sParentMenu + "|" + g_sSubMenu, "");
             llRemoveInventory(llGetScriptName());
         }
-        SetLabel();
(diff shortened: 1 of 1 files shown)


GarvinTwine:
Seems weird to do this twice every time the collar gets rezzed/worn, indeed.
It took me a bit but i finally found the reason for this double draw:
When you do not do the SetLabel(); in state_entry no label will be set when the collar gets passed to a new owner(wearer), which means the label stays as it was from the last owner(wearer), what ever it was.
Might be worth to do that before the reset of owner change instead every time here in state_entry on every rez though!
Lets find the best place for that and add this to the first patch after 4.0 as this is rather a sort of optimization than a bug.

ghost:
the script is reset each time on_rez (line 357-359)
then label text set to default "Open Collar" (line 128)

If the user does not set a new text, the label is drawn in the default text "Open Collar"

ghost:
heck! These buttons are too close!

PR #722, opened 2015-10-30
Wrong order of LMs when changing a prefix caused a false notify with

### LINKED ISSUE #723: old prefix in notification after changing the prefix
thanks to Ruby for reporting this.
fixed in  #722

GarvinTwine:
still the old prefix and not the new one. 
Fixed by sending out the new before doing the notify.
fixes #723


WendyStarfall:
#725

PR #724, opened 2015-11-04
fixed "show restrictions" output

GarvinTwine:
the output missed the first (and if only the only) entry who restricted
what and only showed 
**OpenCollar**: You are being restricted by the following objects
_This collar():_
(same thing already in 3.9xx btw) **OpenCollar - 3.997**: You are being restricted by the following objects
_This collar():_
instead of:
**OpenCollar**: You are being restricted by the following objects
_This collar(vdRestrict): tplm§tploc§tplure§sittp§standtp:_

fixed here.


WendyStarfall:
#726

Issue #729, opened 2015-11-06
auth ranks for restrictions


WendyStarfall:
We don't want to technically deny wearers the ability to "trust" others even if owned with vanilla mode disabled. That also means that a user can add herself as "Trust" access role and potentially override Restrictions set by higher ranked auth roles. Let's try to rank the Restrictions in the RLV Suite by auth. For the RESET command it will mean having to become primary owner only. This change is optional.

PR #730, opened 2015-11-06
for issue #729

GarvinTwine:
added ranks for rlv-restrictions
made RESET and CLEAR ALL owner only functions (wearer can always use
safeword to get released)


WendyStarfall:
#729

Issue #739, opened 2015-11-08
Vanilla buttons is checked when wearer is trusted


GarvinTwine:
When the collar wearer gets added to the trusted list the check box for Vanilla gets checked in the menu which is wrong, and clicking results in a strange message:
_"" cannot be found in owner list_
Also, when added in the right order, the wearer can be in both lists, trusted and owner which should not happen any more.

GarvinTwine:
fixed in #740

PR #740, opened 2015-11-08
fix for issue #739

GarvinTwine:
only set Vanilla check when wearer is owner,
auto-remove wearer from trusted list when they become owner (vanilla
gets enabled).


WendyStarfall:
#739

Issue #750, opened 2015-11-12
a simple method to test the link-set's build integrity


WendyStarfall:
Versions: 4.0.0 to 4.0.11

I've been closely observing others efforts to build the new 4.0 API and I'm happy to have learned that some users figured it out just by digging some through the PRs since mid July or fully on their own.

However, we should add a few basic helpers to prevent mistakes. Zi Ree suggested that link specific scripts could test if they were indeed in the right link. I think this is a good idea and should be considered.

The most obvious build mistake would be falsely linked `com [root]` and `auth [2]` then nothing will work at all. It's also not hard to test `dialog [3]` because if not getting any menus something obviously went wrong. Things get icky though with `RLV [4]` and especially with `settings [5]` as the latter may not be noticed until a user already further downstream realizes RLV is gone or that nothing is ever stored.

I propose we add a simple test if the script is in the correct link and play a sound and OwnerSay if not. :speak_no_evil:

WendyStarfall:
How about adding a `LINK_ALL_OTHERS` command in `com [root]` that would call a test in all Core5.
- `we test integrity`

That way we would avoid spamming as soon as the script had been dropped into a prim.

PS: It's a builder's responsibility to test integrity before starting to deploy their kit and adding this is already more than you would get with most other things. This should add plenty of convenience. :cake:

GarvinTwine:
OK, i made something here which needs adjustments in all corresponding scripts of course:
the command would be as suggested and this would be the output:
`Otto: ga test integrity`
`OpenCollar 4.0: Starting Core5Scripts integrity test...`
`OpenCollar 4.0:`

`Integrity Test Passed!`

or if something is not right:

`Integrity Test Failed:`

`oc_anim not found!`
`oc_couples not found!`
`LinkedPrim Nr. 2 - oc_auth     (correct)`
`LinkedPrim Nr. 5 - oc_settings     (correct)`
`LinkedPrim Nr. 3 - oc_dialog     (correct)`
`LinkedPrim Nr. 4 - oc_rlvsys     (correct)`

or

`Integrity Test Failed:`

`oc_anim not found!`
`oc_couples not found!`
`LinkedPrim Nr. 3 - oc_auth     (WRONG)`
`LinkedPrim Nr. 6 - oc_settings     (WRONG)`
`LinkedPrim Nr. 4 - oc_dialog     (WRONG)`
`LinkedPrim Nr. 5 - oc_rlvsys     (WRONG)`

or 

[...]

PR #753, opened 2015-11-17
added a final remove of listed items before we delete ourselves, just

GarvinTwine:
another try to fight "Old Script" issues


WendyStarfall:
#750

Issue #757, opened 2015-11-21
oc_dialog: Stack-Heap Collision, Script run-time error


ghost:
It happens after several (5-6) consecutive Force Sits
I guess a bug in the Sensor Dialog

WendyStarfall:
I have been able to reproduce this on `4.0.14`

> [10:20] Wendy Starfall: we sit
> [10:21] Wendy's Collar: Sit down, and stay there!
> [10:21] Wendy Starfall: we sit
> [10:21] Wendy's Collar: Sit down, and stay there!
> [10:21] Wendy Starfall: we sit
> [10:21] Wendy's Collar: Sit down, and stay there!
> [10:21] Wendy Starfall: we sit
> [10:21] Wendy's Collar: Sit down, and stay there!
> [10:21] Wendy Starfall: we sit
> [10:21] Wendy's Collar: Sit down, and stay there!
> [10:21] Wendy Starfall: we sit
> [10:21] Wendy's Collar: Wendy's Collar [script:oc_dialog] Script run-time error
> [10:21] Wendy's Collar: Stack-Heap Collision

PR #758, opened 2015-11-21
Fix Bug  #757

ghost:
kID <> kAv typo
This caused unexpected extra DIALOG in UserCommand without the recipient


WendyStarfall:
#757

PR #759, opened 2015-11-22
oc_particle.lsl: untargeted dialogue

ghost:
I do not understand why it is needed.
because LM_SETTING_RESPONSE  returns with empty kMessageID, and it causes unacceptable ConfigureMenu dialogue to 'null_key user'.
in this case, it is meaningless.
I comment it for the future
maybe someone has an idea for this purpose

### DIFF
--- a/src/collar/oc_particle.lsl
+++ b/src/collar/oc_particle.lsl
@@ -169,6 +169,7 @@ Debug(string sStr) {
 */
 
 Dialog(key kID, string sPrompt, list lChoices, list lUtilityButtons, integer iPage, integer iAuth, string sMenuName) {
+    if (kID == "" || kID == NULL_KEY) return; // sanitize NULL_KEYS 
     key kMenuID = llGenerateKey();
     llMessageLinked(LINK_DIALOG, DIALOG, (string)kID + "|" + sPrompt + "|" + (string)iPage + "|" + llDumpList2String(lChoices, "`") + "|" + llDumpList2String(lUtilityButtons, "`") + "|" + (string)iAuth, kMenuID);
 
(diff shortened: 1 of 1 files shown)


GarvinTwine:
It is needed for a remenu when clicking Strict in the Configure menu.
Lets do instead of '//' a if (kMessageID) to prevent null_key being passed to dialog when this message comes from the settings script.
you will find in oc_leash at line 
559
`llMessageLinked(LINK_SET, LM_SETTING_RESPONSE, g_sSettingToken + "strict=1,"+ (string)iAuth, kMessageID);` 
568
`llMessageLinked(LINK_SET, LM_SETTING_RESPONSE, g_sSettingToken + "strict=0,"+ (string)iAuth,kMessageID);`
572
`llMessageLinked(LINK_SET, LM_SETTING_RESPONSE,"strictAuthError="+(string)iAuth,kMessageID)`
this is where a valid key is passed for the menu call in 
oc_particle

ghost:
In this case, it is necessary to check the validity kMessageID, because the oc_settings.lsl returns empty key

GarvinTwine:
that's what i said Romka

ghost:
maybe better need to add check kRCPT for NULL_KEY in oc_dialog.lsl ?
this will prevent the opening of any listens for empty or NULL_KEYs 'recipients'
 #760

GarvinTwine:
i think both...
First of all all scripts should make sure to send only valid dialog LMs, that is the responsibility of the scripts that calls for a dialog.
As seen it happens to be forgotten... (which should be fixed! ) it might be good to have a secure check in dialog itself, too.

PR #760, opened 2015-11-22
oc_dialog.lsl:  sanitize NULL_KEY

ghost:
sanitize NULL_KEY kRCPT to prevent unexpected dialogues


GarvinTwine:
This makes sense still #759 should make sure no null-key dialog is requested.

PR #762, opened 2015-11-26
"Sit down and stay there" does not stay.

ghost:
It seems the command denying unsit was missing and also pointlessly there's a RLV command to stand up before sitting that I've removed.


WendyStarfall:
#763

WendyStarfall:
A solution to this has been provided with 8d2f7da003d4390a06a29cad47515508d7f0c911 and 89b6c71a061d810aa8acf22729853670a9ca06b7. Thanks for the PR!

Issue #763, opened 2015-11-26
Force Sit: Optionally prevent the wearer from standing up.


WendyStarfall:
At the moment the `Force Sit` feature is intended to forcefully seat the avatar.

The collar wearer is encouraged to `stay there` as forced seating doesn't necessarily imply that standing up should be restricted in some way.

During tests we assumed in some versions that if force sat, it should naturally be up to the Owner to unseat. However, feedback largely showed that this was perceived as bug where the collar wearer found herself erroneously stuck when trying to move on. Users felt frustrated and annoyed by the restriction.

If the collar wearer was prevented from standing up by other means in the collar, say through some plugin or terminal command, she must first be unseated in order to be forcefully seated elsewhere.

That brings us to the case where the user logs off or crashes and remains restricted. If so, we will want her to be re-seated on the same spot, but this same spot (a poseball?) might have vanished in the meantime. We will need a timer... and more..

For `4.1.x` next year we should tackle this, but please let's discuss it here and think it through well. :speak_no_evil:

ghost:
Perhaps it could then say in chat in the meantime "Sit down",  or there could be a second menu option for "Sit down and stay there".  That should mitigate complaints,  whilst adding the option for what it seems was intended before.  I can understand the unsit OC function's use now though. 

Perhaps it should redo the sit restraints if they were there before, if there is an easy enough way to check.  I can forsee someone adding an unsit restraint,  force sitting someone,  then sitting them on something else (say they picked the wrong poseball). At this point the original unsit restriction would be undone? Unless I haven't noticed that already coded.

ghost:
I should add,  that having the sit and stay option there alongside sit,  would help sidestep that issue mentioned in the bottom paragraph whilst giving the user more control.

WendyStarfall:
Okay then, here is a little band-aid that improves usability and prevents a situation where the user ends up confused because she forgot that stand up was restricted and then is stuck on the next thing she sat.

The `<prefix> forbid stand` restriction works correctly and is lifted when we receive either `[Get up]` via menu button or `<prefix> stand` via chat command or when we log in.

If standing up is restricted and we are already seated but shall be seated from this spot onto a different spot, there won't be a silent fail now but an informative notify that explains the situation. 

Thanks to @RomkaSwallowtail for experimenting with me and @GarvinTwine for doing how I did just in a way that made it also work!

[...]

PR #776, opened 2015-12-12
Update oc_rlvsuite.lsl

ghost:
add:
 'strict' to Force Sit menu
 'sit back' chat command
force sit on last sit place on login if sit strict


WendyStarfall:
#763

PR #778, opened 2015-12-14
oc_rlvsuite.lsl: automatically apply stand restriction

ghost:
maybe it will be more useful:
when forced sit or sit back, automatically apply stand restriction.
when forced stand - automatically clear stand restriction.

also 'strict' button and chat command 'forbid/allow stand' woks with stand restriction.


ghost:
too early, many people will not understand

PR #779, opened 2015-12-23
Update oc_couples.lsl - force unsit

ghost:
add force unsit before couples


GarvinTwine:
I dont think this is a good idea because it can be used easily to cheat, just hug someone and you are free to stand up though you are meant to keep sat there.

PR #787, opened 2016-01-05
Update oc_remote

ghost:
remove unused fuctions and variables in oc_remote_dialog.lsl

new way for buttons texture, like themes in collar


WendyStarfall:
Sorry, I can't take the card variation.

Yesterday I have tested it and it works fine, however, it makes updating more complicated because the card has to be swapped each time we want to add new buttons variations in patches.

Can I have a PR for `oc_remote_dialog.lsl` only, please?

ghost:
ok, oc_remote_options.lsl return back...

PR #795, opened 2016-01-14
This will bop the api version up by two blips.. yay!

WendyStarfall:
With 4.0 still in pre-release mode we are learning that linking objects in a correct order is apparently too difficult for a great many builders. *cough* So lets after all add a sliver of code in favour of improved flexibility. I signed off on that because I really want to see people being able to link a kitty bell to their collar or just modify it some without bursting into tears after having totally broken it.

hard times for wendy

(Stylization of user having broken her own collar by linking thingy... It's me, alright but you get the picture.)

Our "core5link" scripts will now send their link number to where is relevant and everything will be just fine even if linking not in the correct order. Additionally to that we also have a more reliable way now to upgrade 3.9 API and older collars but that comes later.

We realized already many years ago that downgrading in our case means re-adding bugs or losing important changes. Unfortunately the nature of full perm objects is that everybody can share everything so especially when we released new patches, people asked in the group for a patch, had an even older version they had installed sent to them and then came complaining how the update failed. No more now, nope, nuh uh. Now we check first, thank you very much!

Similar problems appear with totally obsolete Owner HUDs and the likes floating around. "Half of the buttons on your sh*t HUD ain't workin'!" is not an uncommon way for our merry users to approach us when their mate handed them a six years old Owner HUD. We really need a clean slate there for our personalized object channels. That will then totally break old HUDs but it is better the user realizes this don't work anymore with the new stuff, so she can go and pick up a cost-free brand new item, that not only works perfectly but is also more resource friendly, more simple to use and has more feats.

tl;dr then:
- linking and unlinking becomes much easier
- we prevent "downgrading" to broken versions
- a new personalized channel is added

### DIFF
--- a/src/ao/oc_ao_interface.lsl
+++ b/src/ao/oc_ao_interface.lsl
@@ -46,16 +46,18 @@ debug(string sMessage)
 //=
 //= returns      : Channel number to be used
 //===============================================================================
+/*
 integer GetOwnerChannel(key kOwner, integer iOffset) {
     integer iChan = (integer)("0x"+llGetSubString((string)kOwner,2,7)) + iOffset;
     if (iChan > 0) iChan=iChan*(-1);
     if (iChan > -10000) iChan -= 30000;
     return iChan;
 }
-
+*/
 init() {
     g_kObjectID = llGetKey();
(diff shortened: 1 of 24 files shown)


george-black:
I love the picture! :smile_cat:

PR #801, opened 2016-01-16
oc_ao_options.lsl - like oc_remote_options

ghost:
total rework.

it changed a lot, but it works

### DIFF
--- a/src/ao/oc_ao_options.lsl
+++ b/src/ao/oc_ao_options.lsl
@@ -56,6 +56,7 @@ integer SUBMENU = 3002;
 integer DIALOG = -9000;
 integer DIALOG_RESPONSE = -9001;
 integer DIALOG_TIMEOUT = -9002;
+
 //Added for the collar auth system:
 integer CMD_NOAUTH = 0;
 integer CMD_AUTH = 42; //used to send authenticated commands to be executed in the core script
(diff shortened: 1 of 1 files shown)


GarvinTwine:
i havent found where and why yet, but there seems to be somewhere a loop that keeps settings the prim color to black all the time, else this seems to work great

src/ao/oc_ao_options.lsl
```diff
@@ -69,749 +70,315 @@ integer CMD_OWNER = 500;
...
+}
+
+
+default {
+    changed(integer change) {
+        if (change & CHANGED_OWNER) llResetScript();
+        else if (change & CHANGED_LINK) llResetScript();
+        else if (change & CHANGED_COLOR) {
+            DetermineColors(); // -- If we change color because of tint, we need to set the new g_vAOoffcolor!
             DoStatus();
```
GarvinTwine: this causes a loop, color changed, determineColors(), Do Status() color changes again... until all is black

GarvinTwine:
ok, all back i found the (my) issue:
 g_lButtons += llGetLinkPrimitiveParams(i, [PRIM_NAME]);
(my prims where named all "Object" which causes a LINK_SET instead of finding the right prim#
i think we should change this to PRIM_DESC as here the description is already used for the touch event and not the prim name

not that important, i just ran into this issue because i copied the new scripts into the old ao prims which were named that way (lazy me as this is way easier than moving my anims)

ghost:
You can change PRIM_NAME to PRIM_DESC, if necessary

I worked with AO, which gave Wendy

Issue #806, opened 2016-01-17
The new remote and what it looks like out of the box.


WendyStarfall:
Let's make the remote look like this when it is attached for the first time.

![remote_default](https://cloud.githubusercontent.com/assets/1100007/12376883/82a2a1b8-bd09-11e5-9600-2cc88bc59193.png)

PR #807, opened 2016-01-17
Update oc_remote_options

ghost:
for #806

when first attach to HUD upper left


src/remote/oc_remote_options.lsl
```diff
@@ -219,7 +219,7 @@ DefinePosition() {
     integer iPosition = llListFindList(g_lAttachPoints, [llGetAttached()]);
     vector size = llGetScale();
 //  Allows manual repositioning, without resetting it, if needed
-    if (iPosition != g_iSPosition && iPosition != -1) { //do this only when attached to the hud 
+    if (iPosition != g_iSPosition) {
```
GarvinTwine: can we get this change back please, else the hud will move around each time you work on it when rezzed on the ground

ghost: fixed

PR #821, opened 2016-01-20
oc_anim.lsl - is a typo?

ghost:
I believe that this is a typo.
correct me.

### DIFF
--- a/src/collar/oc_anim.lsl
+++ b/src/collar/oc_anim.lsl
@@ -257,7 +257,7 @@ integer SetPosture(integer iOn, key kCommander) {
 }
 
 MessageAOs(string sONOFF){ //send string as "ON"  / "OFF" saves 2 llToUpper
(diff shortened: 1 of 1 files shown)


GarvinTwine:
The first for sure is a typo. the 2nd i guess, too, though these detach events rarely if ever effect anything anyway ;) . lets merge this

ghost:
I have a question:
why duplicated ZHAO-commands into 2 channels: g_iAOChannel and  g_iInterfaceChannel via ATTACHMENT_RESPONSE ?
I know that g_iAOChannel  is used for non-OpenCollar and built-in AO

g_iInterfaceChannel listen in oc_ao_interface.lsl and hanle all commands with AUTH
g_iAOChannel listen in oc_zhao_interface.lsl and hanle only ZHAO commands without AUTH
what for?

GarvinTwine:
this is for future compatibility, the AO uses both channel at the moment for different things, maybe soon we can change that and only listen to the interfaceChannel and retire the AOChannel, but to have both will no need to update the collar scripts if the AO is changed

Issue #831, opened 2016-01-23
Remote HUD "also removes access to collar when removing partner from hud"


GarvinTwine:
This is broken currently and opens up the collar's remove menus for owner and trust list.

Easy solution: remove this feature, no need to try to remove oneself from collars nearby.

Or do it right, but as we dont know which role a remote owner has, we send all 3 (tempowner makes sense, too) and will get 2 errors reported:
OpenCollar: Otto removed from owner list.
OpenCollar: Access denied.
OpenCollar: "garvin twine" is not in trust list.

I personally tend to, remove this feature.

WendyStarfall:
I also say we remove it. This is an artifact-botch to rid of undesired ownership. We are not the collar police. Protecting people by giving them the ability to hunt down the undesired pet to first add it to ones remote and then to bounce oneself off their collar, while they can still just add one back, is just silly. :speak_no_evil:

PR #832, opened 2016-01-23
added a switch back to ALL when removing the active partner from the hud

GarvinTwine:
also commented the removal from collars out for now, needs further
decision and change


WendyStarfall:
#831

PR #836, opened 2016-01-24
get rid of the 2 strided auth lists!

GarvinTwine:
finally we decided to remove the names from the auth lists as we use
anyway only IDs
from now on, no more strided auth lists for owner, trusted or block,
just UUIDs.
added a limiter to loadSettings so our limiits for the auth lists cannot
be cheated by websettings or .settings notecard
Also changed in oc_auth keys to strings as lsl treats them the same but we need them mostly as strings anyway

### DIFF
--- a/src/collar/oc_auth.lsl
+++ b/src/collar/oc_auth.lsl
@@ -21,7 +21,7 @@
 //                    |     .'    ~~~~       \    / :                       //
 //                     \.. /               `. `--' .'                       //
 //                        |                  ~----~                         //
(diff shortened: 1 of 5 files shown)


WendyStarfall:
...and for doing that on a Sunday, in one go, you deserve a year's worth supply of Stiegl.

Cheers and thanks! :beers: 

stiegl-weisse-naturtrueb-bier-sortiment

PS: Too bad we are so underfunded all we get are pics. :disappointed:

PR #844, opened 2016-01-29
stops the installer to play random sounds

GarvinTwine:
random repeats of the sounds where played without the script actually
triggered the sound by llPlaySound commands.
I removed the llPreloadSound on state entry and the issue was gone.
Seems to have been reported once logn time ago but doesnt seem to bother
many at all:
https://jira.secondlife.com/browse/VWR-23660


WendyStarfall:
hehe, thanks

Issue #873, opened 2016-03-09
Leash & strict


ghost:
chat command "strict on/off" causes the "Configure" menu

WendyStarfall:
I forgot why it did that but @GarvinTwine remembered: If we don't re-menu to "Configure", we end up not getting the re-menu when setting strict in the menu checkbox as well. That's why this particular chat command pops the "Configure" menu.

ghost:
it is right, but I think I fixed it

WendyStarfall:
Oh you did!

PS: But it is still strange it wasn't done this way in the first place?

GarvinTwine:
I dont remember why it was done this strange way either...
But if we change this, shouldnt the menu for "AuthError" on chat command also be removed (line 617) ?

``` LSL
            } else if (sToken == "strictAuthError") {
                g_iStrictMode = TRUE;
                ConfigureMenu(kMessageID, (integer)sValue);
            } 
```

We do need the `g_iStrictMode = TRUE;` though

GarvinTwine:
I think I remember now why this is done this "strange" way and to get the menu by chat command was seen as the "lesser bad":
Strict leashing is like leashing Ranked. But the particle script does not know the rank, so what we do here in the menu is:
Someone unchecks the Strict box in the menu... the particle script sets its internal g_iStrictMode = FALSE;
If we do now a remenu right away, the box will be unchecked! Though it might be that the rank was not matched and the leash script reports a "No Access" and strict leashing stays enabled!
To work around that, we wait for the leash script to report the success or failure of unchecking [x]Strict.
I have to test to verify this but I am pretty sure that is the reason of the strange way for the Configure re-menu when hitting the Strict button.
We rather give a "not expected menu" on chat command than a false indication in the re-menu.

ghost:
okay, I will seek another solution

ghost:
try this #876

PR #874, opened 2016-03-09
Leash & strict

ghost:
#873 chat command "strict on/off" causes the "Configure" menu


ghost:
okay, I will seek another solution

Issue #896, opened 2016-03-25
'Smart Strip' should be in the relay.


ghost:
Typically, one reason to use 'smart strip', which replaces @remoutfit with @detachallthis in certain cases, is to help with compatibility in furniture which either does not understand the nature of attachment-based clothing, or tries (unsuccessfully) to be intelligent about what should be removed.

Along with some setup in the RLV folders, smart strip can help alleviate this problem, as it tells RLV to remove the clothing layer and anything else in the same RLV folder. This functionality is provided in the Un/Dress add-on, however it only applies if the stripping is done from the addon itself, making it rather useless in satisfying the use-case in the previous paragraph.

WendyStarfall:
Just something to consider and think about: Maybe `smartstrip` should be stored as a setting? :speak_no_evil:

ghost:
Yes, it should. I didn't know the relay still had that ability. I've added the functionality in #898.

ghost:
As for documentation..

Smart Strip helps alleviate issues encountered with some furniture failing to properly strip an avatar when wearing attached clothing, such as rigged mesh. It does this by associating clothing layers with attachments in your #RLV folders. When smart strip is enabled and a clothing layer is stripped, anything in the same folder with it will be removed as well.

To set it up, create a new folder in your #RLV folder tree and place the attachment(s) (not links, the actual items) you wish to be considered part of that article of clothing into it. Then place a copy of the appropriate 'dummy' clothing layer in the folder as well.

Make sure the entire folder is worn, including the dummy clothing layer. If you plan to wear more than one layer of the same type, you may need to slightly edit the dummy layer, usually its tint, to enable the layers to stack properly.

PR #898, opened 2016-03-29
Add settings integration for SmartStrip in Relay.

ghost:
Adds the smartstrip boolean to the relay settings bitmask that gets sent to and retrieved from oc_settings.


ghost:
Part of ongoing work as per issue #896.

GarvinTwine:
ok i think there is missing something, the value gets saved but not set when loaded from settings

PR #910, opened 2016-04-11
give notify that the outfit feature does not work with RLV (only RLVa)

GarvinTwine:
and to use #folders instead


WendyStarfall:
#914

PR #912, opened 2016-04-11
Use RLVs to adjust pose heights!

GarvinTwine:
Added up and down arrows if RLVa is detected for poses to adjust the
hover height for each pose.
A general hover height can be set without any pose playing which then
will be the one switched back to after a pose's height was adjusted.
settings switch to turn the feature off and set the increment for
up/down:
INTERN=AllowHover~0,0.01
where allowed is 0 or 1 and any sensible float for the increment, if no
float is set, the default of 0.02 will be used.


WendyStarfall:
#913

PR #923, opened 2016-04-20
The final sixified relay plugin.

WendyStarfall:
This is the sixified relay plugin (aka "Das Relais") from the Virtual Disgrace fork.

It also has the standard formatting like all other OpenCollar Six scripts and received a major cosmetic overhaul for the menu which should now be a lot more user friendly. The "Min Mode" feature that raised so many questions from users is now automated, meaning if a higher ranked avatar set something, the wearer can't set it "lower". Chat notifies offer a bit more info for options and "helplessness" will reveal or hide the safeword button along with disabling it if the user chooses to. Relay things are also "spoken by the relay" so to reflect better than relay and collar while in this working together, are really two separate things. If the user doesn't want the relay, she can simply uninstall it with "prefix rm relay". Documentation will be added on the website, smartstrip "dummy layers" will provided embedded with the readme in-world.

Based on @SatomiAhn's masterpiece the authors for the derivative work are @GarvinTwine and I.


george-black:
Super cool :D  Looking forward to trying this (and studying the code).

PR #928, opened 2016-05-03
remove no more needed sit functions

GarvinTwine:
collect menus in one "Stuff" menu to keep the RLV Main menu one sided.


WendyStarfall:
Thanks!

PR #931, opened 2016-05-16
Great AO Idea! but need small fix :)

ghost:
swap stand on / off commands
collar sends commands  "STAND OFF"  when start animation, and "STAND ON"  when stop animation


GarvinTwine:
thanks for the catch! :)

PR #933, opened 2016-05-17
Owners and wearer can (un)lock the AO, unlock of course wearer only if

GarvinTwine:
not an Owner locked it. (Menu cannot be used but to unlock or open the collar menu (for wearer) while locked). Collar on/off command fail silent (unable to sent notify as the collar does not include an ID for these commands)
Removed the auto-unlock for collars safeword, not sure if this makes
sense or not here.

### DIFF
--- a/src/ao/oc_ao.lsl
+++ b/src/ao/oc_ao.lsl
@@ -19,7 +19,7 @@
 //                                          '  `+.;  ;  '      :            //
 //                                          :  '  |    ;       ;-.          //
 //                                          ; '   : :`-:     _.`* ;         //
-//     OpenCollar AO - 160517.2          .*' /  .*' ; .*`- +'  `*'          //
+//     OpenCollar AO - 160518.6          .*' /  .*' ; .*`- +'  `*'          //
 //                                       `*-*   `*-*  `*-*'                 //
 // ------------------------------------------------------------------------ //
 //  Copyright (c) 2008 - 2016 Nandana Singh, Jessenia Mocha, Alexei Maven,  //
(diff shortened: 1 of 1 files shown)


WendyStarfall:
Hi, thank you and I think this was needed very much.

With this the safeword becomes "redundant" to some extent because wearers are now unable to lock themselves out and I see why you feel it could actually be removed altogether. (Of course, they would be able to access the AO via AO Menu in their collar's animation menu but it was still an obstacle.)

The safeword is required because of what happens once you `lock` the AO. It's really a full lockout and that's, not desirable, in arguably most scenarios. The `Sit Anywhere` feature is gone then and the AO can no longer be turned off by the wearer if they were sat on single- or multi-pose furniture.

I'd love to see an AO that is useful. Allowing Owners to access this AO and also load sets etc. is kind of cute and somehow intimate (in a really nerdy way) but "lock out" is, to the best of my judgement of our user base, exceeding what the vast majority really expects and that's: locked = non detachable.

Or a bit more provocative: We cater heavy equipment to a vanilla audience. They will cheer if they get a solid AO, made of only a single script, which they can RLV lock on their HUD to survive outfit changes.

GarvinTwine:
OK, so lets make this a user friendly AO.  #933
The lock feature is a detach-lock now. No more locking out of the AO, it is merely useful to lock it and dont lose it during outfit changes. 
Also changing here that everyone who gets access to the AO can toggle the detach-lock without any ranking included and the AO stay full functional, usable and adjustable while locked.

PR #940, opened 2016-06-21
Update oc_badwords.lsl

ghost:
bug: badwords disappear form Apps menu after collar update

### DIFF
--- a/src/spares/oc_badwords.lsl
+++ b/src/spares/oc_badwords.lsl
@@ -21,7 +21,7 @@
 //                    |     .'    ~~~~       \    / :                       //
 //                     \.. /               `. `--' .'                       //
 //                        |                  ~----~                         //
-//                          Badwords - 160207.1                             //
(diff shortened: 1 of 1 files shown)


GarvinTwine:
I dont see how these changes would change the behaviour to be as button in the Apps menu.

src/spares/oc_badwords.lsl
```diff
@@ -183,7 +183,7 @@ ParseAnimList(string sStr) {
             g_lAnims = llDeleteSubList(g_lAnims,i,i);
             if (sTest == "~shock") g_iDefaultAnim = TRUE;
         }
-    } while (i);
```
ghost, line 186: this causes an infinite loop when the animation list is empty. change to (i>0)

it is possible if the request for the list of animations occurs earlier than the animation script is ready to go.
so I moved the request animations list at the end of the load settings

other changes are "cosmetic"

PR #945, opened 2016-07-05
new themes

ghost:
I propose a new format for .themеs notecard:
ElementName ~ texture=texture_id ~ color=color_vector ~ shiny=shiny_type ~ glow=glow_level

I think it is more convenient, informative and flexible.

Element parameters can be arranged in any order, like:
ElementName ~ color=color_vector ~ texture=texture_id ~ glow=glow_level ~ shiny=shiny_type

Ability to use 2 and more lines for same element, like:
ElementName ~ color=color_vector ~ texture=texture_id
ElementName ~ glow=glow_level ~ shiny=shiny_type

Unnecessary element parameter can be skipped, like:
ElementName ~ texture=texture_id ~ shiny=shiny_type

(It may be the same method is suitable for particles in .themes?)

### DIFF
--- a/src/collar/.themes
+++ b/src/collar/.themes
@@ -1,3 +1,3 @@
 [ ThemeName ]
-ElementName~<texture_id>~<color_vector>~<shiny_type>~<glow_level>
-Ring~89556747-24cb-43ed-920b-47caed15465f~<1,1,1>~none/low/medium/high/specular~none/low/medium/high/veryHigh
+ElementName~texture=<texture_id>~color=<color_vector>~shiny=<shiny_type>~glow=<glow_level>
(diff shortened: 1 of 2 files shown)


GarvinTwine:
I think this is in general not a bad idea.
But there is one big issue:
If the themes script is updated by the general patch installer, the old themes note cards will not work any more. As every collar design has its own .themes note card, I do not see a way to also replace the current .themes note cards with new ones...
So updating the themes script would result first of all in failure of all collars which have the old style themes note cards and only new collars with already updated note cards would work.

In general only the style of the note card is changed, which is a note card used by designers and not end users... Those who actually use it by now, have understood how they work and their collars would need themes note card updates as well...
So, though the idea is not bad, the change would cause way more issues than the benefit of a "more convenient, informative and flexible" note card style.

ghost:
Yes, this is a problem :(
and why this idea has not occurred to me before? before the release of 6.0 :(

PR #947, opened 2016-07-20
new themes format

ghost:
another attempt 
new format for .themеs notecard:
ElementName ~ texture=texture_id ~ color=color_vector ~ shiny=shiny_type ~ glow=glow_level

I think it is more convenient, informative and flexible.

Element parameters can be arranged in any order, like:
ElementName ~ color=color_vector ~ texture=texture_id ~ glow=glow_level ~ shiny=shiny_type

Ability to use 2 and more lines for same element, like:
ElementName ~ color=color_vector ~ texture=texture_id
ElementName ~ glow=glow_level ~ shiny=shiny_type

Unnecessary element parameter can be skipped, like:
ElementName ~ texture=texture_id ~ shiny=shiny_type

Also can use ALL for all elements, like:
ALL ~ color=<1,1,1> ~ shiny=medium
ElementName ~ texture=texture_id ~ shiny=none

(It may be the same method is suitable for particles in .themes?)

PS: old format also can be used for compatibility.

### DIFF
--- a/src/collar/oc_themes.lsl
+++ b/src/collar/oc_themes.lsl
@@ -19,7 +19,7 @@
 //                                          '  `+.;  ;  '      :            //
 //                                          :  '  |    ;       ;-.          //
 //                                          ; '   : :`-:     _.`* ;         //
-//           Themes - 160629.1           .*' /  .*' ; .*`- +'  `*'          //
+//           Themes - 160720.1           .*' /  .*' ; .*`- +'  `*'          //
 //                                       `*-*   `*-*  `*-*'                 //
 // ------------------------------------------------------------------------ //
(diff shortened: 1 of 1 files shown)


WendyStarfall:
I think it is good and I like it. I wonder about memory use, we should profile it.

ghost:
it additionally consumes about 528 bytes

WendyStarfall:
I really want this too. This a much better method for designers.

It would be too risky to ship it with 6.2.5 though and we should wait for a 6.3.x.

In the meantime, designers who want to use this method, can simply load the script from here.

WendyStarfall:
#948

Issue #948, opened 2016-07-29
A new format for themes


WendyStarfall:
**Version:** `6.3.x`
**Script/s:** `oc_themes`

A couple years ago, during the discussion for `OpenCollar - paint` @RomkaSwallowtail created a very cool plugin. It was called `OpenCollar - styles` and would allow the user to turn all settings from appearance modules into a single "style" that could be printed, pasted to a card and conveniently loaded.

However, the big issue we always had with our appearance modules were the sheer amount of them.

If all were installed, we had eight such modules and the user would require every, single one of them if she wanted to be able to fully customize her accessory. That's actually the best and most logical approach.

Unfortunately, the way Second Life® became during the last few years, a very large part of the user base were led to believe that _"the more scripts an accessory had, the worse it would be"_. Toot! Toot! :hear_no_evil:

OpenCollar even took a kind of "unixy approach" in its beginnings ([DOTADIW](http://dotadiw.com/) and some of [Raymond's Rules](http://tinyletter.com/programmingphilosophy/letters/eric-raymond-s-17-unix-rules)) which was fun and great at the time and while it is sad that trends rejected this, it gave us new challenges that resulted in the highly optimized and hyper-tidy OpenCollar Six device that we have today and...

![get-to-the-point](https://cloud.githubusercontent.com/assets/1100007/17258439/2d1049d8-55c7-11e6-9bf7-143903231c85.jpg)

:speak_no_evil: Okay! Long story short, the new `oc_themes` takes care of all appearance modules, in a single script, where the composition of dedicated themes is created by the artist. The configuration for themes is a bit confusing though, so @RomkaSwallowtail proposed an alternative configuration style in #947.

> @RomkaSwallowtail commented on Jul 20, 2016, 4:14 PM GMT+2
> 
> another attempt
> new format for .themеs notecard:
> ElementName ~ texture=texture_id ~ color=color_vector ~ shiny=shiny_type ~ glow=glow_level
> 
> I think it is more convenient, informative and flexible.
> 
> Element parameters can be arranged in any order, like:
> ElementName ~ color=color_vector ~ texture=texture_id ~ glow=glow_level ~ shiny=shiny_type
> 
> Ability to use 2 and more lines for same element, like:
> ElementName ~ color=color_vector ~ texture=texture_id
> ElementName ~ glow=glow_level ~ shiny=shiny_type
> 
> Unnecessary element parameter can be skipped, like:
[...]

PR #950, opened 2016-08-05
oc_themes.lsl - Wendy has found bug: All the textures list for every element in Looks - Texture menu

ghost:
fix textures list for 'Element~Texture,uuid'


WendyStarfall:
#948

PR #951, opened 2016-08-06
oc_themes.lsl - very little optimization?

ghost:
there is no reason to keep the variables used only once
replace constants


WendyStarfall:
#948

Issue #944, opened 2016-07-03
The *touch* feature is misleading with current artwork


WendyStarfall:
Some of the notifies in Themes, when using the Looks variant, were not correct and I'm removing the button for the *Touch* feature that @SatomiAhn wrote during the 3.8 phase but keep the logic intact as this is actually really cool and might come in handy for other developers who mess with our source. 

However, many assemblies have slightly larger touch parts now that cover most of the artwork, so this is not really useful unless the user wanted to tint or texture her invisible Storage or Animator cylinders.

PR #953, opened 2016-08-07
oc_themes.lsl - more optimisations

ghost:
#948

136, 137: remove unused variables

241: forgotten "Touch" in Elements menu

337-338: It saves up to 1 KB script memory

540: use constant "global_DeviceType" instead  g_sGlobalToken+"DeviceType"

### DIFF
--- a/src/collar/oc_themes.lsl
+++ b/src/collar/oc_themes.lsl
@@ -133,8 +133,8 @@ integer g_iStylesNotecardLine;
 integer g_iLeashParticle;
 integer g_iLooks;
 
-//string g_sSettingToken = "themes_";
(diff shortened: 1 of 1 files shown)


WendyStarfall:
The **Touch** feature is a bit of a problem with modern assemblies. Every fragile and thin looking design is covered by the "touch prims" around the collar. Those are usually the cylinders that hold settings and animations. Many users already had big problems _touching_ the collar on the avatar's neck. That eases it.

Cabochon Hertz
The large cylinders around the design ease touchability for menu, grabbing, capture etc..

The logic stays intact for whoever wants to modify and use it. (#944)

Issue #958, opened 2016-09-10
no relay involved


KarinKorpov:
Uuhhmm, let try to understand the problem

One of my pets with collar version 6.3.3 has the restriction unsit, she going sit on something and relog. After the relog grabs the collar not back on the object where she belongs and she got go to wander around. The restriction "unsit" stay there till the timer is done.... 

Conclusion, i think
The collar goesn't back to the UUID what the collar in mind have, that she has with force sit.

GarvinTwine:
The title of the issue is already the answer to your problem here.

The behaviour is as designed, the unsit restriction does not allow to stand up. But as long as you do not sit it as has "no effect". So neither if you relog. It is not a auto-resit enforcement!

Where in your case this might be a desired behaviour, in many others it might not!
After all this is unsit and not capture on a device!

If you want a permanent forced sit on a specified furniture, the furniture can use rlv and the collars relay to do so. That is meant for capturing on a device.

As this effect is already covered with RLV-relay using devices, I think the way it is now makes all happy.

PS: 6.3.0 is the current release, not sure if this is a typo or some other non-official OpenCollar fork:
http://www.opencollar.at/updates.html

KarinKorpov:
The item has not a lock option or something else about RLV.... It's was a normal chair where the pet going on sit and I did strict in "force sit"... when I this doing with other restrains, works this very well..... Is this what you mean?

P.s. Yes i mean 6.3.0 version

GarvinTwine:
I dont know about other restrains, but if you like them, use them by all means!

The collar's unsit feature is build in within reason into the Force-Sit feature:
It is meant to keep one seated after forcing them to sit. In this case, the collar also knows where (on what UUID) the wearer was forced on and will force it back on that UUID after relog.
Without using force sit, this whole things get confusing as it will work only under certain circumstances.
So the design was meant like it is, with thought how it was made.

WendyStarfall:
Merged https://github.com/VirtualDisgrace/opencollar/commit/3e810f958e35833b050d94a2b3fbe01c12c9cd1b into `VirtualDisgrace/opencollar:alt.sit` for future reference.

There is an alternative take on this with https://github.com/VirtualDisgrace/opencollar/pull/961/commits/48a90737f97dd80bee6134d45ae7c9a7b3c2f5e9 in #961 that I find also good but couldn't merge.

PR #960, opened 2016-09-11
Update oc_rlvsuite.lsl - force sit back after relog

ghost:
#958
additionally force-sit when owner turn on "strict", if sub is already sitting.
it makes force-sit back after sub relog.

### DIFF
--- a/src/collar/oc_rlvsuite.lsl
+++ b/src/collar/oc_rlvsuite.lsl
@@ -778,6 +778,7 @@ default {
         else if (iChan == g_iSitRLV) {
             if ( (key)sMsg != NULL_KEY) {
                 g_kLastForcedSeat = (key)sMsg;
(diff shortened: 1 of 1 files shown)


WendyStarfall:
This would mean the user can be forcefully kept without being forcefully sat.

The code is correct and works but I don't think this is where we want to take the device as it might lead to further confusion about a feature most users are already comfortable with as it is. I'm sorry. :disappointed:

This might lead to further discussion about whether to use the relay at all for such things or the collar and, again, what most users are already comfortable with is the knowledge that such things are usually part of the furniture that makes use of the relay plugin.

It would be nice if you keep the branch with this particular edit intact in case somebody prefers this style. 

Thanks in any case!

KarinKorpov:
After a the script change with this change, works again.. Thanks for this!

PR #961, opened 2016-09-12
Update oc_rlvsuite.lsl - strict and force sitting

ghost:
#958 - another way

apply 'unsit=n' restriction only if already force-sitting


WendyStarfall:
Oops! I was going to also merge this into the alt.sit branch and caused a conflict.

Seems like @KarinKorpov was satisfied with #960 already. It's all good. 🙊

Issue #962, opened 2016-09-20
AO: Some animation states need to be translated


WendyStarfall:
**Version:** `OpenCollar AO 6.1.3 (160706.1)`

**Problem:** Our default AO sets and the resulting custom made AO sets still use the old ZHAOII names for animation states. Many values are the same in `llSetAnimationOverride`, however some differ by a missing space in between words and again some animation states have a completely different name altogether.

**Solution:** We translate these values in the script. I'm sure there are more correct/elegant methods to do this and maybe we should have a translator of some kind that reads existing ZHAOII and Oracul cards to spit out a properly formatted and optimized (no empty lines) card but I think this wouldn't be very user friendly.

Keep in mind that the user would have to do several complicated steps on her own with a translator device. She would have to take an AO set card, drop it into an object, let the object process the card, copy the print and paste it into a new card. That sounds like not really a big deal to us but arranging a custom AO set is a major technical achievement for non-technical users. :speak_no_evil: 

~~If we create such a translator device, I suggest to make some kind of "AO Set Maker" that guides the user through the process step by step, along the lines: _"Hold down Ctrl, select the walking animation in your inventory, drag and drop the walking animation into the AO Hud....Well done! You dropped the walking animation into the AO Hud....Hold down Ctrl, select the sitting animation in your inventory, drag and drop your sitting animation into the AO Hud...Yay! You dropped your sitting animation into the AO Hud...etc"_~~

I came up with a much better solution in [the next post](https://github.com/VirtualDisgrace/opencollar/issues/962#issuecomment-249131093). 🍰

WendyStarfall:
Okay then. Here is a much more user friendly solution to this:

In #965 I'm adding all the known state values from Oracul, AX and RP overrider configurations.

This already gives us the ability to simply unpack an Oracul AO for example and shove it all over into the OpenCollar AO. I've been testing this with a rich featured AX type, the "Maidenly Grace", and was positively surprised by still having 12% of memory available in the `oc_ao.lsl` script.

To complete with maximal user friendliness, I propose an installer script with the following usage flow:
1. Rez the Oracul/AX/RP AO on the ground.
2. Rez the OpenCollar AO next to it on the ground.
3. Drop the `oc_ao_loader` into the Oracul/AX/RP AO.
4. Check for a dialog menu to pop up.
5. Click one of the 125cm/155cm/185cm/215cm heights in the dialog menu.
6. An installation begins. Wait for it to finish.
[...]

PR #966, opened 2016-09-24
Add support for multi load page and oracul loader

GarvinTwine:
#962


WendyStarfall (approved):
Works flawlessly. Free memory varies in between 6% to 10% with hyper-crammed set cards.

PR #971, opened 2016-10-01
Perform an additional operation on safewords to trim spaces

stawberri:
Back when my safeword actually did something, I noticed that pressing alt+enter on Firestorm produced a message like `(( SAFEWORD ))` while OpenCollar expected a message like `((SAFEWORD))`. This seemed rather inconvenient in a situation that would merit needing to use a safeword.

My implementation converts the wearer's message into a list and and then pulls the first element from it only if the list is one element long. This works out because the safeword setting command does a similar thing where it pulls only one word from the message.

Now that I think about it, parsing every message someone sends sounds like a lot of overhead, so if you like this idea, I could look into adding an extra check if the first or last character of the string at that point is a space. If neither is a space, then this whole list thing is completely unnecessary. 😄

### DIFF
--- a/src/collar/oc_com.lsl
+++ b/src/collar/oc_com.lsl
@@ -416,6 +416,9 @@ default {
             if (llGetSubString(sw, 0, 3) == "/me ") sw = llGetSubString(sw, 4, -1);
             if (llGetSubString(sw, 0, 1) == "((" && llGetSubString(sw, -2, -1) == "))") sw = llGetSubString(sw, 2, -3);
             if (llSubStringIndex(sw, g_sPrefix)==0) sw = llGetSubString(sw, llStringLength(g_sPrefix), -1);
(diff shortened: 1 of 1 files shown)


GarvinTwine (changes requested):
Nice catch, though i think Firestorm should change their OOC as they show it without spaces in the Preference-Window ;).
But LSL has nice function to Trim Spaces:
llTrimString(string,TRIM_STRING); that added to the if condition where the "(())" are found and all should be fine, the /me case is already handled anyway above that.

stawberri:
Okay, I did that, and it seems to work!

My `/me` case _isn't_ actually handled though, because my example had multiple spaces in it (like `/me    SAFEWORD`). I have less of a rationale for needing to support that though. 😅

WendyStarfall (approved):
Tested and works well!

GarvinTwine (approved):

WendyStarfall:
#974

Issue #972, opened 2016-10-01
RLV Restrictions get ignored if they (their string) are first part of an existing restriction of the same ID


GarvinTwine:
@TonyaSouther pointed this out to me:

>  Tonya Souther: I found a problem in rlvsys: if you issue a command that's a substring of a command you've already issued, it's silently dropped.

If multiple restrictions by the same ID are send to rlvsys, it can happen that one gets ignored.

rlvsys saves restrictions each ID sends in a string:
§restriction1§restriction2§restriction3...
When the same ID (usually a plugin) sends new restrictions it gets checked if they are already set, no need to re-add  and save something that is already there.
Restrictions get added one by one by the script, even if sent in one command.
This check has a little flaw in special cases:
`if (!(sSrcRestr==sBehav || ~llSubStringIndex(sSrcRestr,"§"+sBehav) || ~llSubStringIndex(sSrcRestr,sBehav+"§"))`

The issue occurred when detach restrictions for the HUD were sent:
detach:top left=n,detach:top right=n,detach:top=n
detach:bottom left=n,detach:bottom right=n,detach:bottom=n

Here the last of the three gets ignored due to:
`~llSubStringIndex("§detach:top left","§detach:top") = -1`

Where the script says "hey i know you _(partially at least)_, dont want you again".
Not sure if there are any other possible restriction where this happens.

I think this can be easily fixed by changing the if statement to:
`if (!~llListFindList(llParseString2List(sSrcRestr,["§"],[]),[sBehav]))`
as the whole strings are compared here and not only partial.

PR #973, opened 2016-10-01
change the check for existing restrictions when trying to add a new one

GarvinTwine:
issue #972

### DIFF
--- a/src/collar/oc_rlvsys.lsl
+++ b/src/collar/oc_rlvsys.lsl
@@ -251,7 +251,7 @@ AddRestriction(key kID, string sBehav) {
     string sSrcRestr = llList2String(g_lRestrictions,iSource+1);
     //Debug("AddRestriction 2.1");
     //if (!(sSrcRestr==sBehav || ~llSubStringIndex(sSrcRestr,"§"+sBehav) || ~llSubStringIndex(sSrcRestr,sBehav+"§")) ) {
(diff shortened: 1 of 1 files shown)


stawberri:
Would it be more efficient to add a `§` to the beginning and end of `sSrcRestr` and check for the existence of `§restriction§`?

GarvinTwine:
That would require first to check what other effects such a change would have to other parts of this script and eventual other scripts that might use this. 
I know the print out of "show restrictions" does and would should show as:
This collar (vdTerminal): §detach:top left§§detach:top§
instead of now:
This collar (vdTerminal): detach:top left§detach:top

Cosmetics I guess but well...
I am not sure if either is more or less effective at all here, it's about one if statement that executes 2 ll functions, i guess the llSubStringIndex is a bit faster then to parse and ListFindList but then this might have other changes followed that negotiate the win again...

So I think this fix for a possibly years old little bug nobody even noticed till these days, this is fine

stawberri:
Um, I think I might not have been clear enough! I meant that those symbols could be added during the check. 😅 

``` lsl
if (!~llSubStringIndex("§"+sSrcRestr+"§","§"+sBehav+"§")) {
```

GarvinTwine:
I definitely misunderstood, so here you go :)

WendyStarfall (approved):
Tested with various sources, this is good to go!

Issue #970, opened 2016-09-30
Installers allow you to attempt two installations at once, resulting in the loss of all settings.


stawberri:
# Steps to reproduce
1. Rez a collar.
2. Configure some setting on your collar. For example, `<prefix> safeword EXAMPLE`.
3. Rez two copies of the OpenCollar Patch.
4. Start one of the updates.
5. While the other update has reached at least 25% but hasn't completed yet, start the other.
# Results
- The first patch will become stuck.
- Your collar will lose all of its settings. Check this by typing `<prefix> safeword` and see that it has reset.
# Real world scenario:

In my steps, I suggested using two patches because they're easy to obtain and take forever to install. When I actually discovered this bug, I was trying to update and install an app at the same time.
# My best guess of the cause
## Patch getting stuck

The first installer's `oc_update_shim` gets replaced by the second's.
## Settings loss

`oc_update_shim` seems to save current settings, update scripts, and then restore the settings. The installer that runs second may be reading the settings when the scripts no longer have them, and then restoring those empty settings onto the collar.
# Potential solution

At some point in the beginning of the installation process, check if there is already an installation happening, and refuse to complete the installation.

GarvinTwine:
We already try to avoid this scenario when you start the update/installation from the collar by the "Update" button or command.
But right, if someone does it like you did, things don't go well.
At least for updates this could be solved by simply keeping the oc_sys script (which gets alwasy as first installed) not running until the update is done, so no 2nd installation can get started as this script provides the pin to enable to load scripts.
But this doesnt solve it if someone rezzes 5 different Apps and an update and just clicks them quickly one after the other... Here we can add a scan (same as the sys script does when the Update command is issued) to see if there are other installers/updaters around and if... we simply do nothing or even delete ourselves.
I just fear we might at some point run into one issue we won't be able to fix:

> A common mistake that people make when trying to design something completely foolproof is to underestimate the ingenuity of complete fools.

Douglas Adams

stawberri:
In that case, maybe we could try something simpler, like adding a message to the installation prompt warning people not to do two installations at once? There's still room for error, but if a "WARNING" message was there I (probably) wouldn't have made this mistake.

WendyStarfall:
I'll go out on a limb here and tell you all that I'm less than convinced of this fix/change.

[...]

PR #975, opened 2016-10-02
check on rez for other installers

GarvinTwine:
and dont allow installations to initiate when one is found.
#970

### DIFF
--- a/src/installer/oc_installer_sys.lsl
+++ b/src/installer/oc_installer_sys.lsl
@@ -21,7 +21,7 @@
 //                    |     .'    ~~~~       \    / :                       //
 //                     \.. /               `. `--' .'                       //
 //                        |                  ~----~                         //
-//                       Installer System - 161002.2                        //
+//                       Installer System - 161003.1                        //
 // ------------------------------------------------------------------------ //
 //  Copyright (c) 2011 - 2016 Nandana Singh, Satomi Ahn, DrakeSystem,       //
 //  Wendy Starfall, littlemousy, Romka Swallowtail, Garvin Twine et al.     //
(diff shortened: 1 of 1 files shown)


WendyStarfall (changes requested):
I've played around a bit in-world and thought that we maybe should use a notify function, similar to as we already do in other devices. Right now our installers all have very long names. Something like:

``` LSL
Notify(key g_kUser, string sMsg) {
    llSetObjectName("Installer");
    llRegionSayTo(g_kUser,0,sMsg);
    llSetObjectName(g_sObjectName);
}
```

We can also consider to lend the user a hand and print a SLURL to where the other installer is hiding. I've messed with that too but only had this dorky function:

``` LSL
string GetSLURL() {
    string sMaps="http://maps.secondlife.com/secondlife";
    string sRegion=llGetRegionName();
    vector vLocal=llGetLocalPos();
    string sX=(string)llRound(vLocal.x);
    string sY=(string)llRound(vLocal.y);
    string sZ=(string)llRound(vLocal.z);
    return (sMaps+"/"+llEscapeURL(sRegion)+"/"+sX+"/"+sY+"/"+sZ);
}
```

Maybe it's better to just do a NameURI for the other installer device. Those hyperlinks display world coordinates with `/inspect` anyway. Helping the user to find hidden installers might be important if we take the highlander solution. In a worst case, the user fails to find the other installer and gets very frustrated. 🙊

GarvinTwine:
I will name it Say(string) as this goes only to the owner anyway and so differs from our Notify function.
There's from the way of "doing it" no real difference if printing out a URI or map link as both need the same formatted SURL (for objects this differs from agents).

WendyStarfall (approved):

WendyStarfall:
#982

PR #978, opened 2016-10-06
Fix typo preventing intern setting responses from being read

stawberri:
It turns out that while pose hover heights were being saved to the intern setting, they actually weren't being loaded properly because of a typo.


GarvinTwine:
looks like i really missed a freaking "_" here... thanks!

ghost:
also take a look at ~heightscalars, g_lAnimScalars, and everything associated with it: it is not used anywhere...
maybe it is reserved for the future?

GarvinTwine:
I think you are right Romka, it is rather forgotten from the past than reserved for the future :)

WendyStarfall:
#983

PR #985, opened 2016-10-13
Update oc_shocker.lsl

ghost:
fix ParseAnimList endless loop and cosmetic


george-black:
rawr rawr K&R braces!!! <3

PR #1001, opened 2016-11-08
CMD_SAFEWORD

ghost:
I think we just forget it?


WendyStarfall:
We don't need to call the safeword at this spot!

The relay safeword and the global safeword are treated a bit differently. Relay only safewords can be used for people who don't want to spam their owners with safeword notifies and will affect only the relay. 🙊

ghost:
I think global safeword must issue relay safeword too, for inform capture device to release victim.
but not vice versa

WendyStarfall:
That's what is already happening also without the edit.

ghost:
no, when safeword oc_rlvsys send RLV_CLEAR linkmessage, bur oc_relay don't handle it!
so, relay will not know what safeword happened and don't sent to capture device '!release' command,
and device capture victim again

WendyStarfall:
This kind of behaviour was persistent for roughly half a decade. It's what everyone knows and expects when the global safeword is used. Contraptions may even re-capture. I don't understand why these tiny edits always have to happen a couple days after major releases even though there are weeks of time in between.

ghost:
okay, you right

PR #1012, opened 2017-03-05
Adding Extra Exception

DrakeSystem:
Adding @accepttpoffer to the list of Exceptions,
also split Sendim into Sendim and Startim (was only one checkbox)


WendyStarfall:
EDIT: That is bound to confuse some users but it really belongs there. I think that it could be added to a special branch that was previously mentioned, granted that users of this special branch will explain one another what those exceptions actually mean and what impact they have on experience. Unfortunately I don't have the time and peace of mind to take care of something like this at the moment.

PR #1013, opened 2017-03-10
Leash and Stealth

ghost:
I think it will be better to hide the leash when the collar is invisible ?

### DIFF
--- a/src/collar/oc_particle.lsl
+++ b/src/collar/oc_particle.lsl
@@ -126,7 +126,7 @@ key g_kLeashToPoint;
 key g_kParticleTarget;
 integer g_iLeasherInRange;
 integer g_iAwayCounter;
-
+integer g_iGlobalShow;
 integer g_iLeashActive;
 integer g_iTurnMode;
 integer g_iStrictMode;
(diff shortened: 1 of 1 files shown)


stawberri:
While this would be convenient, I think it might still be desirable sometimes to have a visible leash with an invisible collar.

WendyStarfall:
We thought about doing that when creating 4.0 but then quickly realized that a healthy amount of users would wear a stealthed OpenCollar 4.0 above unscripted accessory and ended up not stealthing the leash.

It's still worth considering I think because of the name we chose for hiding the collar, "stealth", it implies going unnoticed which is something a lot of people in our subculture do in their daily lives.

If we expand there it would open possibilities to auto-stealth when in PG zones or under other set conditions. Maybe hiding a collar and stealthing a collar should be two different things.

In any case, I don't volunteer to develop this or explain it to users. Haha! 🙊

ghost:
Need another way...

WendyStarfall:
There should be a difference in between "hide" and "stealth".

I liked your change @RomkaSwallowtail !

PR #1021, opened 2017-04-01
Display range limit status when listing access details

stawberri:
There is a command to limit / unlimit the access range of collars for group and public users, but I couldn't find any location (other than a settings print) that would actually display the current state of this setting. I think this might put owners at risk of accidentally setting someone's collar to sim range and forgetting about it.

After trying a bunch of different things, I settled on on displaying this information as an extra row in the access list, only when it is applicable and set to simwide access:

* I decided on having it show up only when it is applicable because it's not really necessary to know that a collar has simwide access when the people who would need it can't access the collar in the first place.
* I decided on having nothing extra show at all when the collar is set to limited range access, because this is the default, and displaying this would likely cause confusion for inexperienced owners.

### DIFF
--- a/src/collar/oc_auth.lsl
+++ b/src/collar/oc_auth.lsl
@@ -458,11 +458,17 @@ UserCommand(integer iNum, string sStr, key kID, integer iRemenu) { // here iNum:
                 sOutput += "\n" + NameURI(llList2String(g_lBlock, --iLength));
             if (sOutput) llMessageLinked(LINK_DIALOG,NOTIFY,"0"+"Blocked: "+sOutput,kID);
             //if (g_sGroupName) llMessageLinked(LINK_DIALOG,NOTIFY,"0"+"Group: "+g_sGroupName,kID);
-            if (g_kGroup) llMessageLinked(LINK_DIALOG,NOTIFY,"0"+"Group: secondlife:///app/group/"+(string)g_kGroup+"/about",kID);
+            if (g_kGroup) {
+                sOutput = "simwide";
+                if (g_iLimitRange) sOutput="chat range";
+                llMessageLinked(LINK_DIALOG,NOTIFY,"0"+"secondlife:///app/group/"+(string)g_kGroup+"/about: "+sOutput,kID);
+            }
             sOutput="closed";
-            if (g_iOpenAccess) sOutput="open";
+            if (g_iOpenAccess) {
+                sOutput = "simwide";
+                if (g_iLimitRange) sOutput="chat range";
+            }
             llMessageLinked(LINK_DIALOG,NOTIFY,"0"+"Public Access: "+ sOutput,kID);
-            if (!g_iLimitRange) if(g_iGroupEnabled || g_iOpenAccess) llMessageLinked(LINK_DIALOG,NOTIFY,"0"+"Access range: simwide",kID);
         }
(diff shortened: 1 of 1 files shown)


WendyStarfall:
We had it in the "disgraced version" in the owner list print similarly to how you did there. I think too that it is more user-friendly to state it in this print but I forgot why it wasn't added to the OpenCollar. 🙈 

Maybe someone else knows, I really can't remember!

GarvinTwine:
Not sure if displayed  **only** if not set and **only** when on group or public access is active, doesn’t confuse users again :). "Now it's there, now it's not..." sort of thing.
If then, I think it should be displayed what ever setting is active. 
Just my 2 cents though. :)

stawberri:
I guess having it only appear in such a specific case might be a little confusing, but I still think that having it show up all the time would be a bad idea.

```
Public Access: open
Public Access Range: chat range
```

```
Group: Some Group Name
Public Access: closed
Group Access Range: simwide
```

```
Group: Some Group Name
Public Access: open
Public Access Range: simwide
```

These three are good, but~

```
Public Access: closed
Public Access Range: chat range
```

This one is a little confusing. People might complain, "I turned off my public access, but it still says my access range or whatever is chat range! I'm mad!"

So I think it would be best if this final case would just not display a range?

```
Public Access: closed
```

GarvinTwine:
I rather thought of a one time print:

` Access Range: sim wide / chat range`

As it effects also Owners and Trusted as long they do not use a Remote.

stawberri:
I don't really understand what you mean by "one time print." 😅

But um, I think shortening the message to just "Access Range" might risk making people think that it means their owners need to stay in chat range for commands to work.

GarvinTwine:
I guess that is all why it isn't displayed.
I find it myself to confusing...

stawberri:
I guess so. I decided to try something else, though! 😓

Group and public disabled:
image

Public enabled, limitrange on:
image

Public enabled, limitrange off:
image

Public & group enabled, limitrange off:
image

Group enabled, public disabled, limitrange off
image

Is this better? 😅

GarvinTwine:
I like that :)

WendyStarfall:
I remember why this was taken out:

The limitrange always confused users. Nobody really knew what it meant and people came up with their own explanations. I've made up my mind and think that it is sufficient to reflect the limitrange state in the settings print only and not in the Access List print.

This is going to sound obnoxious but explaining to users over and over what the "range" stuff means, clarifying that "simwide" means within the same sim, telling that chat range means 20 meter etc.. isn't something I want to do. I'll get even more obnoxious and claim that "most users are better off with limited range" because, let's face it, a lot of them will get extra-butthurt if people "touch their collars without asking first" and exactly those users will run around with unlimited range, I promise you that, so they can get artificially worked up about something and find likeminded llamas in the OC group to communally wine how horrible everyone on SL is for clicking at the accessory on their neck from afar... 🙊 

I'm content with ``AUTH=limitrange~0`` and while I think that the Access List print is where it really belonged as human readable info, I believe that we shouldn't put it there (for obnoxious reasons).

EDIT: A special branch for this and some other features that required more direct user explanation was maybe a bit too ambitious an idea. I sadly don't have the time and peace of mind to approach this at the moment.

stawberri:
Hmm, okay! I guess that makes sense. :3

PR #1022, opened 2017-04-01
Remove g_sGroupName and auth_groupname

stawberri:
While I was making my other pull request, I noticed that collars still fetch and save the name of your group and display its fetched name when group access is enabled. However, the access list displays your group's name using its URI. This seems like a strange discrepancy, so I removed group name fetching completely.


GarvinTwine:
I like that, no idea why we forgot to remove it in the first place. 
So good catch and idea!

WendyStarfall:
#1026

PR #1023, opened 2017-04-05
Drag landmark to OpenCollar remote to teleport partners there

stawberri:
Since bookmarks aren't included in the collars anymore, it might be a good idea to make it even easier for people to teleport their partners around. The OpenCollar remote is conveniently located at a fixed location on your screen, so it makes a _really_ easy to use drag-and-drop target for inventory items.

This isn't an entirely new idea so I won't claim that I came up with it, but I wrote up a small feature that allows you to drag and drop a landmark onto your remote to teleport your partners to the location of your landmark.

image

I was a little lazy and wanted to write as little code as possible (which conveniently results in a more consistent experience for users), but the downside of doing this is that processing a landmark and preparing for the teleport is slow. To "fix," this, I had the remote give two progress messages while this process is happening. They're spaced out just right to give a nice countdown feeling after you drop a landmark. 😊

image

It's probably also important to mention that I changed the "follow your partners" message that you get. I noticed that my lazy landmark destination scripting resulted in teleport urls with out-of-sim offsets. These _do_ seem to work in Firestorm, but they just show up as secondlife: urls instead of the nicely presented location links that users are used to. I was worried that this might scare some people, so I wrapped the entire thing in a link and replaced it with a simple message about following your partners, with no indication of where the link takes you (you should know, after all).

image

Here's that ugly link you can send yourself in chat to see if it works (or displays nicely) for you:

```
secondlife:///app/teleport/Bondage/-49282/17534/1234
```

I just realized I misspelled processing.

### DIFF
--- a/src/remote/oc_remote_bookmarks.lsl
+++ b/src/remote/oc_remote_bookmarks.lsl
@@ -66,6 +66,8 @@ list   g_lDestinations_Slurls         = []; //Destination list direct from stati
 list   g_lVolatile_Destinations       = []; //These are in memory preferences that are not yet saved into the notecard
 list   g_lVolatile_Slurls             = []; //These are in memory preferences that are not yet saved into the notecard
 key    g_kRequestHandle               = NULL_KEY; //Sim Request Handle to convert global coordinates
+key    g_kLandmarkData                = NULL_KEY; //Landmark data request dataserver key
+string g_sLandmarkName                = ""; //Current landmark inventory item name
 vector g_vLocalPos                    = ZERO_VECTOR;
 string g_sRegion;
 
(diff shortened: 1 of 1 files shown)


GarvinTwine:
Nice idea.
The issue with the strange coordinates is about global and local coordinates I think.
The TeleportTo() function expects local coordinates but gets in this case global ones.
So g_vLocalPos gets filled with global coordinates. For the collar tp the global coordinates are used anyway so it doesnt matter but for the display and app/teleport they should be local again ...
But to verify this, I have to test and debug it first.

WendyStarfall:
EDIT: I found this to be a really cute idea and wanted to see it shipped, but unfortunately there isn't anyone testing and I don't have the free time and the peace of mind at the moment to do it all on my own.

PR #1024, opened 2017-04-14
Leash & 'Stealth' (take two)

ghost:
Added the option to hide the leash if the collar is invisible
(Probably need another name for this option?)


stawberri:
If this is what I think it is, how about calling it "Auto Hide"? 🤔 

I haven't actually tested if it fits on a button or not, but it's more clear than just "Hide."

ghost:
"Auto Hide"? who not :thinking:
What others will say about this?

WendyStarfall:
EDIT: The feature makes lots of sense but it will be complicated to explain to users. Sadly I don't have the time and peace of mind these days to take care of a special branch that would be better suited for it.

PR #1025, opened 2017-05-23
Fix antislide run animation detection

stawberri:
The current method of detecting whether or not an antislide run animation exists actually determines whether or not a full permission run animation exists. I changed it to use `llGetInventoryType`, which works even for animations that are not full perm.

There is a similar check for the built-in `~run` animation that I didn't change, because there's no reason for `~run` not to be full perms.


WendyStarfall:
#1027

Issue #1045, opened 2017-07-28
oc_resizer - presets save/load


Davwyn:
Feature Request:

Would it be possible to add an option to the resizer script, or create an addon for the collar to be able to print out all the current position/rotation settings the collar is currently in that you can dump into a notecard and store in the collar. Then using the menus, recall those settings from the respective notecard to have the collar quickly and easily change size/position appropriate to whatever outfit the avatar is wearing?

This would be a fantastic feature for pets that have radically different shapes and sizes like going from human to an actual dog avatar etc.

Perhaps borrow code from the oc_settings script for the export/import notecard function.

WendyStarfall:
With the stealth logic separated from ``oc_themes`` such a feature could be added there but a completely new plugin would be better. Coders who are competent with such things are @RomkaSwallowtail and if memory gets tight @littlemousy.

ghost:
#1047 
Maybe it needs polishing, but it already works

WendyStarfall:
Thanks @RomkaSwallowtail   ❤️

ghost:
#1048 
another way - standalone plugin

PR #1048, opened 2017-08-19
Create oc_presets.lsl

ghost:
standalone, based on resizer, plugin for save and restore size/position

Maybe it needs polishing, but it already works


WendyStarfall (approved):

WendyStarfall:
#1045

PR #1049, opened 2017-08-20
resizer2 and presets small updates

ghost:
сorrect preset deletion,
add delete settings when script remove


WendyStarfall:
#1045

PR #1050, opened 2017-09-20
Particles in titler

manonev:
Needs a new prim with appropriate desc (and should degrade gracefully in its absence).  Not hooked into menu system yet; has to be done with # title particle <KEY>.  And that's another thing it doesn't have yet: shorthand names for some standard particles (I've made and found a few; someone with better graphic skills could do good things).

This is cooler than it sounds: the particles are not flying around; they're stationary graphics hanging in the air.  So they could be words like hovertext, but with cool graphics and coloring and design and everything.  One notable issue: since the particles have to be *at* a prim, there has to be a prim overhead for them, which means a prim linked to but fairly far from the collar, so it moves pretty widely with small movements of the head.

See what you think,

### DIFF
--- a/src/collar/oc_titler.lsl
+++ b/src/collar/oc_titler.lsl
@@ -147,15 +147,18 @@ ShowHideText() {
         if (g_sText == "") g_iOn = FALSE;
         llSetLinkPrimitiveParamsFast(g_iTextPrim, [PRIM_TEXT,g_sText,g_vColor,(float)g_iOn, PRIM_SIZE,g_vPrimScale, PRIM_SLICE,<0.490,0.51,0.0>]);
     }
-    if (g_iPartPrim >0) {
-        // Might as well move it...
-        llSetLinkPrimitiveParamsFast(g_iPartPrim,
-                                     [PRIM_POS_LOCAL,
-                                      g_vPartOffset + <0.,0, g_vPrimScale.z>]);
+    if (g_iPartPrim > 0) {
         if (g_sParticle == "") {
           llLinkParticleSystem(g_iPartPrim,[]);
+          if (g_iPartPrim == g_iTextPrim) {
+            llSetLinkPrimitiveParamsFast(g_iTextPrim,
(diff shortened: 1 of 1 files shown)


WendyStarfall:
I like it and thanks for the contribution. There is a withering concept called ``titler++`` somewhere on my graveyard on bitbucket that also uses particles over classic hovertext that was never finished. We will have to see how it works out with client-side lag and use the regular FloatText element, which should be no big deal.

GarvinTwine:
Sorry I didnt get to reply to this earlier. 
I like the idea and it could be added.

I see the main challenge here to enable and allow the users to add their own textures, else it will be a nice feature but with only prepared few textures to use.

As this is really just in the first developing stages, I would suggest to continue in an own branch before we start merging into this one.

GarvinTwine (changes requested):
I see no reason why not using the already existing FloatTextPrim for this, as long as only one of the features can be used, either float text or particles, there should be no conflict at all, where having to add a new prim will make this only work for new collars and not in hundred thousands already out there.

manonev:
I've been working on this some, and have gotten it to a point more worth considering.  I have a small collection of textures in a menu, as well as the option to paste a texture UUID into a textbox, but yes, it really should be something you can add to a little more easily than pasting UUIDs.

I'll do some testing; probably should be possible to drop textures into the collar and use those, perhaps with a prefix or something in case there are other textures for other purposes.  Possible downside to that is proliferation of copies of the "standard" ones, if they're distributed with the collars.  Or maybe some standard ones with UUIDs hardcoded, plus textures dropped in, etc...

I was using a separate prim because I wasn't thinking that we necessarily should allow only one at a time, but perhaps it would be okay.  Or maybe even better, maybe I could make it smart enough to use a dedicated prim IF one was available...

I'll see if I can do some/all of these things, and push my branch soon so you can see what you think.  Thanks for looking!

WendyStarfall:
This would be a perfectly fine, new plugin too by the way.

Issue #865, opened 2016-02-27
Interface Channel: Avoid duplicated channels.


WendyStarfall:
**Version:** `6.0.1 (160224.100000)`

**Problem:** If the 7th character in an agent's id is a 0, the result of the personalized channel could be a duplicate of the HUD channel in which case we would leave the `listen()` function prematurely.

**Solution:** The interface channel could either be generated differently or we just increment there by 1.

This is on me for rushing reduced complexity in Six-com. To my defense, only 1 out of 6,219 users would run into this problem when messing with the new interface stuff. It is improbable but should be changed.

Thanks @GarvinTwine for spotting and figuring this issue out at all. :beer:

PR #1051, opened 2017-09-30
Update oc_com.lsl

ghost:
Meaningless code that is not executed under the condition "if (iChan == g_iHUDChan)"
The same code is below

maybe it was a typo...


GarvinTwine:
It seems meaningless but it is a fix for a special condition that can occur:
issue  #865

ghost:
I'm a stupid cat :))

PR #1053, opened 2017-10-01
oc_com.lsl listen channels optimisation

ghost:
I think this is a way to avoid double code and this will work with the same hud/interface channels...

listen algorithm:
1 - check and hadle for LockMeister (and return if do it)
2 - check and handle for HUD channel
3 - check and handle for Interface channel
4 - check and handle for chat/private channel and safeword

### DIFF
--- a/src/collar/oc_com.lsl
+++ b/src/collar/oc_com.lsl
@@ -416,27 +416,6 @@ default {
     }
 
     listen(integer iChan, string sName, key kID, string sMsg) {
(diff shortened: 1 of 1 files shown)


GarvinTwine:
Looks good to me :)
When reading the code again i noticed that we use quite often llGetOwnerKey in the listen event
Let's do that only once and store the result in kOwnerID after the LockMeister stuff?

ghost:
like this?
it's a good idea

PR #1069, opened 2017-12-13
 Fix graph title to support multiple words.

george-black:
Now we properly rebuild the title from lParams so extra words aren't lost.

### DIFF
--- a/src/collar/oc_titler.lsl
+++ b/src/collar/oc_titler.lsl
@@ -239,7 +239,7 @@ UserCommand(integer iAuth, string sStr, key kAv) {
             } else {
                 g_iOn = TRUE;
                 if (sCommand == "graph") {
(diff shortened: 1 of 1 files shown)


src/collar/oc_titler.lsl
```diff
@@ -239,7 +239,7 @@ UserCommand(integer iAuth, string sStr, key kAv) {
             } else {
                 g_iOn = TRUE;
                 if (sCommand == "graph") {
```
GarvinTwine, line 241: I would suggest to test here if this is meant to be a graph by:
`if (sCommand == "graph" && ( sAction == "0" || (integer)sAction) {`

Else the change works fine!

GarvinTwine (approved):
see comment in code

GarvinTwine (commented):
Sorry but after testing this....
I think it is better without the integer check

WendyStarfall:
#1070

